PR #17, opened 2019-01-15
fix(helm) use the stable helm version

hutchic:
the stable helm repository has been updated to handle Kong 1.0 so lets use that instead of a fork


hutchic:
handled in a different PR

PR #25, opened 2019-02-20
fix(dependency) add libyaml dependency

hutchic:
Kong 1.1.0 is going to require libyaml to be installed

All four `next` builds fail
capture

Now they pass
capture2

Needed to fix daily builds of [next](https://travis-ci.org/Kong/kong/builds/496104400)


Makefile
```diff
@@ -30,6 +30,8 @@ RESTY_VERSION ?= `grep RESTY_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk
 RESTY_LUAROCKS_VERSION ?= `grep RESTY_LUAROCKS_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_OPENSSL_VERSION ?= `grep RESTY_OPENSSL_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_PCRE_VERSION ?= `grep RESTY_PCRE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
+LIBYAML_VERSION ?= 0.2.1
+LYAML_VERSION ?= 6.2.3
```
hutchic: Will snag these values from `KONG_SOURCE_LOCATION` after next merges into master but for the time being just going to set the defaults here

PR #27, opened 2019-03-03
feat(development) add the ability to setup a Kong development environment

hutchic:
Per the readme updates to test this run:
```
make development
```

and proceed with normal development tasks double check anything else needs to be called out in the readme / anything that's missing etc

NB: first build will take time subsequent runs are heavily cached (~20 min vs ~2min)


bungle:
I get this:
```
Error: ./kong/cmd/start.lua:50: nginx: [error] init_by_lua error: ./kong/init.lua:402: error loading plugin schemas: on plugin 'myplugin': myplugin plugin is enabled but not installed
```

was that something that was expected to happen?

hutchic:
damn git submodules 😠 

@bungle when cloning a repo with submodules `git clone --recursive` or `git clone && git submodule init`

bungle:
@hutchic, yes that was it!

jeveleth:
Tested hitting `:8001/`. Got the expected response. Would it be worth adding `git submodule update --init --recursive` as a step in the README?

hutchic:
@jeveleth I nuked the submodule in favour of just having the kong-plugin code live in this repository

PR #37, opened 2019-04-01
feat(src) src release to bintray

hutchic:
Tested locally by building / releasing Kong 1.1.1 from source

```
KONG_VERSION=1.1.1 RESTY_IMAGE_BASE=src make release-kong
... SNIP ...
kong-src
generic
src
1.1.1
kong-1.1.1.tar.gz
kong-src artifact upload: [status = 201] {"message":"success"}


hishamhm:
LGTM — it picks the same .tar.gz from github.com as kong-distributions does, so should have identical results.

PR #92, opened 2019-06-20
feat(redhat 8) 

hutchic:
Redhat is renaming their 7 docker image and has released the base docker image for 8


Makefile
```diff
@@ -151,11 +151,17 @@ build-kong:
...
 	docker pull registry.access.redhat.com/rhel${RESTY_IMAGE_TAG}
 	docker tag registry.access.redhat.com/rhel${RESTY_IMAGE_TAG} rhel:${RESTY_IMAGE_TAG}
+else ifeq ($(RESTY_IMAGE_BASE),rhel)
+	docker pull registry.access.redhat.com/ubi${RESTY_IMAGE_TAG}/ubi
+	docker tag registry.access.redhat.com/ubi${RESTY_IMAGE_TAG}/ubi rhel:${RESTY_IMAGE_TAG}
+endif
+ifeq ($(RESTY_IMAGE_BASE),rhel)
 	PACKAGE_TYPE=rpm
 	@docker build -f Dockerfile.$(PACKAGE_TYPE) \
+	--no-cache \
```
hutchic: cache'd RHEL systems cause issues because they think they're registered with RedHat when in fact they are not

Dockerfile.rpm
```diff
@@ -10,13 +10,15 @@ ARG REDHAT_USERNAME
 ARG REDHAT_PASSWORD
 ARG RHEL="false"
 
-RUN if [ "$RHEL" = "true" ] ; then subscription-manager register --username ${REDHAT_USERNAME} --password ${REDHAT_PASSWORD} --auto-attach ; fi
-RUN if [ "$RHEL" = "true" ] ; then yum-config-manager --enable 'rhel-*-server-optional-rpms' ; fi
+RUN if [ "$RHEL" = "true" ] ; then subscription-manager register --username ${REDHAT_USERNAME} --password ${REDHAT_PASSWORD} ; fi
+RUN if [ "$RHEL" = "true" ] ; then subscription-manager refresh ; fi
+RUN if [ "$RHEL" = "true" ] ; then subscription-manager attach --auto ; fi
+RUN if [ "$RHEL" = "true" ] ; then command -v yum-config-manager && yum-config-manager --enable 'rhel-*-server-optional-rpms' || true ; fi
```
hutchic: RHEL8 doesn't have `yum-config-manager` and doesn't need it so check if the command exists before running it

Dockerfile.rpm
```diff
@@ -10,13 +10,15 @@ ARG REDHAT_USERNAME
...
 
-RUN yum -y install wget tar readline-devel pcre-devel openssl-devel gcc curl unzip \
-  zlib-devel make gcc gcc-c++ bzip2 patch perl m4 git coreutils which
+RUN yum -y --skip-broken install wget tar readline-devel pcre-devel openssl-devel gcc curl unzip \
+  zlib-devel make gcc gcc-c++ bzip2 patch perl m4 git coreutils
 
-RUN which realpath || /bin/bash -c "curl -sSLO https://ftp.tu-chemnitz.de/pub/linux/dag/redhat/el6/en/x86_64/rpmforge/RPMS/realpath-1.17-1.el6.rf.x86_64.rpm && rpm -i realpath-1.17-1.el6.rf.x86_64.rpm"
+RUN command -v realpath || /bin/bash -c "curl -sSLO https://ftp.tu-chemnitz.de/pub/linux/dag/redhat/el6/en/x86_64/rpmforge/RPMS/realpath-1.17-1.el6.rf.x86_64.rpm && rpm -i realpath-1.17-1.el6.rf.x86_64.rpm"
```
hutchic: Stack Overflow pointed out that `command -v` is readily available everywhere while `which` is not

PR #103, opened 2019-07-08
fix(*) removing sudo in Makefile

subnetmarco:
Removing `sudo` commands from the `Makefile` that mess with everybody's home folder.


hutchic:
this has been resolved. All CI dependency installation has been moved to dedicated bash scripts that are not part of the default Make tasks

PR #118, opened 2019-08-16
fix(src) source packaging and release fix

hutchic:
src builds are mostly a noop as the github handles everything. This change makes almost all make tasks a noop and adds a travis-ci check that we can "build" kong source. The travis-ci check doesn't build anything but it would have caught that we goofed the source builds


Makefile
```diff
@@ -83,10 +83,16 @@ BUILD_TOOLS_SHA=$$(cd openresty-build-tools/ && git rev-parse --short HEAD)
...
 setup-ci:
+ifneq ($(RESTY_IMAGE_BASE),src)
 	.ci/setup_ci.sh
-	.ci/setup_kind.sh
+	$(MAKE) setup-tests
+	$(MAKE) setup-build
+endif
 
 setup-build:
+ifeq ($(RESTY_IMAGE_BASE),src)
```
hishamhm: I assume that RESTY_IMAGE_BASE being `src` implies that BUILDX is false, is that correct?

hishamhm: Ah, I see that it's the case here:

https://github.com/Kong/kong-build-tools/blob/6333ebe170315d7d2bf8fbd746338e8af546756c/Makefile#L57

(though perhaps that line should have been `BUILDX=false` without the `?` there, since this is an explicit conditional override?)

hutchic: BUILDX is defaulted to false currently because we don't require ARM architecture so no point setting it up to not use it and just tear it down in the end.

I'll update this to be an `if / else` otherwise it may error when we eventaully do ARM builds 👍

hutchic:
```
make setup-build
nothing to be done
BUILDX=true make setup-build
nothing to be done
```

adjusted and tested locally

Makefile
```diff
@@ -101,6 +106,9 @@ ifeq ($(BUILDX),true)
 endif
 
 cleanup_build:
+ifeq ($(RESTY_IMAGE_BASE),src)
+	@echo "nothing to be done"
+endif
```
hishamhm: same `else` needed here?

PR #123, opened 2019-09-03
fix(lyaml) allow the lyaml version to be modified

hutchic:
@hishamhm pointed out that kong-build-tools isn't consuming the environment variable necessary to change the lyaml version

I think there's better ways we should be instally lyaml but that refactor is a little out of scope for the time I have available / not a high enough priority

ref: https://github.com/Kong/kong/pull/4926


hishamhm:
@hutchic `lyaml` is a regular Kong dependency (just like the bunch of other rocks listed in our rockspec). `libyaml` is a system library dependency that needs to be installed by kong-build-tools in advance. It's a similar relationship to `luaossl` and `openssl`. I think we should not single out `lyaml` in the tooling, and just let the rockspec take care of its version.

hutchic:
agreed we should but for reasons that's been lost to history (and my forgetful brain) I think we found we had to isolate it out. I'll try to PR the removal of installing lyaml in kong-build-tools and will trust the test suite to refresh my memory if it is in fact still needed

hutchic:
closed in lieu of https://github.com/Kong/kong-build-tools/pull/124

PR #176, opened 2019-12-04
feat(systemd) include systemd files with our packages

hutchic:
Allows Kong to be controled via systemd where possible

### DIFF
--- a/Dockerfile.package
+++ b/Dockerfile.package
@@ -9,6 +9,7 @@ FROM kong/fpm:0.0.1 as FPM
 
 COPY --from=KONG /tmp/build /tmp/build
 COPY fpm-entrypoint.sh /fpm-entrypoint.sh
+COPY after-install.sh /after-install.sh
 COPY .rpmmacros /root/.rpmmacros
 ARG PRIVATE_KEY_FILE
 COPY ${PRIVATE_KEY_FILE} /kong.private.asc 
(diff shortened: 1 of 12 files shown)


Dockerfile.package
```diff
@@ -22,6 +23,9 @@ ARG KONG_PACKAGE_NAME
 ARG KONG_CONFLICTS
 ARG BUILDPLATFORM
 
+RUN mkdir -p /tmp/build/lib/systemd/system/
+COPY kong.service /tmp/build/lib/systemd/system/kong.service
```
gszr: Not a blocker: would `/etc/kong` be a better location for this file, similarly to the template configuration file that we ship?

hutchic: (my understanding is) we're putting the default Kong `kong.service` file in `/lib/systemd/system` but then advise the users to customize Kong via `/etc/systemd/system/kong.service` file (ie https://docs.konghq.com/enterprise/1.3-x/systemd/#customize-the-kong-enterprise-unit-file )

FWIW I've very liberally stole most f what you see from https://github.com/Kong/kong-distributions/pull/263 and next sprint will need to PR doc changes that resemble. I have no strong feelings but keeping the two syncronized seems ideal

gszr: It seems misleading to place a configuration file template in the `lib` directory - typically meant to contain shared and/or static libraries. This is, in fact, subject of a Linux standard - the "Filesystem Hierarchy Standard" . But, not a blocker.

PR #201, opened 2020-01-10
fix(base) add missing go-pluginserver build arg

gszr:
edit: failing because Kong/kong master does not have the pluginserver version in its .requirements file.


gszr:
Included in #207

PR #225, opened 2020-02-19
fix(kong-tests) some kong tests need the postgres container to trust them

hutchic:


gszr:
Does this require a sister PR on kong (bump)?

hutchic:
yes it will

PR #236, opened 2020-03-10
fix(ci) use our official dockerfiles for testing

hutchic:
using the official dockerfiles for testing de-dupes so we're certain we're testing what we're releasing. It also makes daily docker images available for publishing


Makefile
```diff
@@ -37,6 +37,7 @@ KONG_GMP_VERSION ?= `grep KONG_GMP_VERSION $(KONG_SOURCE_LOCATION)/.requirements
 KONG_NETTLE_VERSION ?= `grep KONG_NETTLE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 OPENRESTY_PATCHES ?= 1
 LIBYAML_VERSION ?= 0.2.1
+DOCKER_KONG_VERSION ?= chore/remove-builder-pattern
```
hutchic: todo: merge this and then pin this to `master` I guess

Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 pipeline {
     agent none
     environment {
-        KONG_SOURCE = "next"
+        KONG_SOURCE = "master"
```
hutchic: pending the htest fix making it to `next`

Makefile
```diff
@@ -310,7 +311,8 @@ ifneq ($(RESTY_IMAGE_BASE),src)
 	KONG_ADMIN_URI="http://$(TEST_HOST):$(TEST_ADMIN_PORT)" \
 	KONG_PROXY_URI="http://$(TEST_HOST):$(TEST_PROXY_PORT)" \
 	TEST_COMPOSE_PATH=$(TEST_COMPOSE_PATH) \
-	KONG_BASE_IMAGE_NAME=$(DOCKER_REPOSITORY):openresty-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)-$(DOCKER_OPENRESTY_SUFFIX) \
+	DOCKER_GO_BUILDER=$(DOCKER_REPOSITORY):openresty-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)-$(DOCKER_OPENRESTY_SUFFIX) \
+	KONG_BASE_IMAGE_NAME=$(KONG_TEST_IMAGE_NAME) \
```
gszr: I believe this is no longer used.

test/tests/03-go-plugins/up.sh
```diff
@@ -12,6 +12,7 @@ docker cp go-plugin-builder-container:/go-plugins/go-hello.so .
 docker rm -f go-plugin-builder-container
 rm -rf go-plugins
 
+echo $KONG_BASE_IMAGE_NAME
```
gszr: This can be nuked

Makefile
```diff
@@ -348,8 +351,10 @@ endif
 cleanup-tests:
 ifneq ($(RESTY_IMAGE_BASE),src)
 	docker-compose -f test/kong-tests-compose.yaml down
+	docker-compose -f test/kong-tests-compose.yaml rm -f
+	docker-compose -f docker-compose.yml down
```
gszr: Is `docker-compose.yml` used anywhere? Not sure this down and rm belong here in this target.

Makefile
hutchic: `DOCKER_GO_BUILDER` is the openresty image we build go in
`KONG_BASE_IMAGE_NAME` is the Kong docker image built from our docker-kong Dockerfile

the `03-go-plugins/up.sh` builds the go plugin using one then uses the other to cp the go plugin into and test it

Makefile
hutchic: It is used and there is a fair bit of duplication between the two docker-compose files but I'll tackle that in a different PR. I tried merging them in this one and it was very challenging

Makefile
gszr: I might be missing something, but `KONG_BASE_IMAGE_NAME` isn't actually used anywhere, is it? I can only see one reference to it, and it is an `echo` that could be removed.

hutchic: yup I see what happened I thought `KONG_BASE_IMAGE_NAME==KONG_TEST_IMAGE_NAME`

test/tests/03-go-plugins/up.sh
gszr: @hutchic leftover

PR #243, opened 2020-03-19
fix(logrotate) replace CR line terminators by LF

WALL-E:
logrotate will thrown error when a `<CR>` character is encountered.


hutchic:
thanks for that 💯

PR #244, opened 2020-03-20
split all golang into a previous stage

javierguerragiraldez:
this allows the final image to copy only the compiled go-pluginserver.  Later, using `--target
go-phase` we can build a separate image with the exact compiler and environment used to compile
the pluginserver.


Makefile
```diff
@@ -154,6 +154,23 @@ endif
...
+else
+	$(CACHE_COMMAND) $(DOCKER_REPOSITORY):golang-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)-$(DOCKER_KONG_SUFFIX) || \
+	( $(MAKE) build-base ; \
+	$(DOCKER_COMMAND) -f dockerfiles/Dockerfile.golang \
+	--build-arg KONG_GO_PLUGINSERVER_VERSION=$(KONG_GO_PLUGINSERVER_VERSION) \
+	--build-arg RESTY_IMAGE_TAG="$(RESTY_IMAGE_TAG)" \
+	--build-arg RESTY_IMAGE_BASE=$(RESTY_IMAGE_BASE) \
+	--build-arg DOCKER_REPOSITORY=$(DOCKER_REPOSITORY) \
+	--build-arg DOCKER_BASE_SUFFIX=$(DOCKER_BASE_SUFFIX) \
+	--build-arg DEBUG=$(DEBUG) \
```
hutchic: I think this isn't used

```suggestion
```

Makefile
```diff
@@ -154,6 +154,23 @@ endif
...
+	-t $(DOCKER_REPOSITORY):golang-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)-$(DOCKER_KONG_SUFFIX) . )
+	-$(UPDATE_CACHE_COMMAND) $(DOCKER_REPOSITORY):golang-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)-$(DOCKER_KONG_SUFFIX)
```
hutchic: the golang docker image will need to be regenerated if:
- the kong-build-tools sha changes
- the KONG_GO_PLUGINSERVER_VERSION variable changes

I think define a
```
DOCKER_GO_SUFFIX=${BUILD_TOOLS_SHA}-${KONG_GO_PLUGINSERVER_VERSION}${CACHE_BUSTER}
```
around line 85 and use it as the cache key (?)

javierguerragiraldez: done

hutchic:
This looks good. I think the only final bit would be pushing the docker image correct?

Should be able to place that in https://github.com/Kong/kong-build-tools/blob/master/release-kong.sh I believe

PR #273, opened 2020-05-14
feat(nginx-module) add the ability to pin the kong nginx module

hutchic:
pin the kong nginx-module

### DIFF
--- a/build-openresty.sh
+++ b/build-openresty.sh
@@ -34,6 +34,11 @@ then
   KONG_NGX_BUILD_ARGS="--debug"
 fi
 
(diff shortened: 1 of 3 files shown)


build-openresty.sh
```diff
@@ -34,6 +34,11 @@ then
   KONG_NGX_BUILD_ARGS="--debug"
 fi
 
+if [ -z "$KONG_NGINX_MODULE" ]
+then
+  KONG_NGINX_MODULE="master"
```
hutchic, line 39: should we maybe use `0.0.6` as a default?

PR #274, opened 2020-05-15
feat(kong-ngx-build) add OpenSSL FIPS mode support

thibaultcha:
Based on instructions from
https://www.openssl.org/docs/fips/UserGuide-2.0.pdf.

Only OpenSSL 1.0.1 and 1.0.2 are supported by the OpenSSL FIPS module.

* New option `--with-openssl-fips` to enable FIPS mode.
* OpenSSL FIPS module version is hard-coded given it is unlikely to be
  something users care to customize.
* We also double-check that FIPS symbols are included in the final
  libcrypto dylib as a sanity check.

See [FTI-1574](https://konghq.atlassian.net/browse/FTI-1574)

### DIFF
--- a/openresty-build-tools/kong-ngx-build
+++ b/openresty-build-tools/kong-ngx-build
@@ -61,6 +61,10 @@ main() {
         OPENSSL_SHA=$2
         shift 2
         ;;
+      --with-openssl-fips)
+        OPENSSL_FIPS_VER=2.0.16
+        shift 1
+        ;;
       --luarocks)
         LUAROCKS_VER=$2
         shift 2
@@ -140,6 +144,13 @@ main() {
     fatal "OpenSSL version can not be empty"
   fi
 
+  if [[ -n "$OPENSSL_FIPS_VER" ]] && \
+       ! version_eq $OPENSSL_VER 1.0.1 && \
+       ! version_eq $OPENSSL_VER 1.0.2 ];
+  then
+    fatal "OpenSSL FIPS mode is only supported for OpenSSL 1.0.1 and 1.0.2"
+  fi
+
   # retrieve DIST info of DIST-specific patches
 
   if [ -f /etc/os-release ]; then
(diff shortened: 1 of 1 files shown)


openresty-build-tools/kong-ngx-build
```diff
@@ -61,6 +61,10 @@ main() {
         OPENSSL_SHA=$2
         shift 2
         ;;
+      --with-openssl-fips)
+        OPENSSL_FIPS_VER=2.0.16
```
hutchic, line 65: I suspect we want this to eventually be pinned in `kong/.requirements` which would look something like change this line to `OPENSSL_FIPS_VER=$2`

and pull the variable in the Makefile and send it through to kong-ngx-build with edits:
https://github.com/Kong/kong-build-tools/blob/4.3.6/Makefile#L37
https://github.com/Kong/kong-build-tools/blob/4.3.6/dockerfiles/Dockerfile.openresty#L68
https://github.com/Kong/kong-build-tools/blob/4.3.6/build-openresty.sh#L47

openresty-build-tools/kong-ngx-build
```diff
@@ -140,6 +144,13 @@ main() {
     fatal "OpenSSL version can not be empty"
   fi
 
+  if [[ -n "$OPENSSL_FIPS_VER" ]] && \
+       ! version_eq $OPENSSL_VER 1.0.1 && \
+       ! version_eq $OPENSSL_VER 1.0.2 ];
+  then
+    fatal "OpenSSL FIPS mode is only supported for OpenSSL 1.0.1 and 1.0.2"
```
hutchic, line 151: should we fatal or just unset the varialbe?

thibaultcha: Certainly not unset the variable and ignore it.

openresty-build-tools/kong-ngx-build
thibaultcha: The commit and PR explain why this is hard-coded.

hutchic:
I think this will need a companion PR in kong-distributions to effect Kong-ee

eskerda:
> I think this will need a companion PR in kong-distributions to effect Kong-ee

On it. We need to add the flag here https://github.com/Kong/kong-distributions/blob/354786a111c041ec750e75933e81c0983ba7ce30/kong-images/build-kong.sh#L291-L300 , I guess conditionally

gszr:
hey @thibaultcha, pinging to check if this PR is still relevant

thibaultcha:
See https://konghq.atlassian.net/browse/FTI-1379 https://konghq.atlassian.net/browse/FTI-1574

fffonion:
Closing this as boringssl is used for FIPS mode, we will go back to openssl 3.0 but that will be different compile routine.

PR #299, opened 2020-06-22
chore(*) do not bundled go-pluginserver with kong builds

gszr:
make sure to squash and merge


test/Dockerfile.test
```diff
@@ -28,11 +29,21 @@ RUN ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime && \
...
+ENV PATH $GOPATH/bin:$PATH
+
+RUN go version ; \
+	mkdir /gps ; cd /gps ; \
+	go mod init go-pluginserver ; \
+	go get -d -v github.com/Kong/go-pluginserver@${KONG_GO_PLUGINSERVER_VERSION} ; \
+	go install -ldflags="-s -w -X main.version=${KONG_GO_PLUGINSERVER_VERSION}" ... ; \
+	cp /tmp/go/bin/go-pluginserver /usr/local/bin/ ;\
+	cd ; rm -r /gps; \
+	go-pluginserver --version
```
hutchic, line 46: if we also removed the go tests do we require golang and go-pluginserver?

test/Dockerfile.test
```diff
@@ -2,16 +2,17 @@ ARG RESTY_IMAGE_BASE="ubuntu"
...
 ARG DOCKER_REPOSITORY
 
-FROM ${DOCKER_REPOSITORY}:go-plugin-tool-${RESTY_IMAGE_BASE}-${RESTY_IMAGE_TAG}-${DOCKER_GO_SUFFIX} as GOLANG
-
 FROM ${DOCKER_REPOSITORY}:test-${DOCKER_OPENRESTY_SUFFIX}
 
 ENV PATH=$PATH:/kong/bin:/usr/local/openresty/bin/:/usr/local/kong/bin/:/usr/local/openresty/nginx/sbin/
 ENV LUA_PATH=/kong/?.lua;/kong/?/init.lua;/root/.luarocks/share/lua/5.1/?.lua;/root/.luarocks/share/lua/5.1/?/init.lua;/usr/local/share/lua/5.1/?.lua;/usr/local/share/lua/5.1/?/init.lua;./?.lua;/usr/local/openresty/luajit/share/luajit-2.1.0-beta3/?.lua;/usr/local/openresty/luajit/share/lua/5.1/?.lua;/usr/local/openresty/luajit/share/lua/5.1/?/init.lua
 ENV LUA_CPATH=/root/.luarocks/lib/lua/5.1/?.so;/usr/local/lib/lua/5.1/?.so;./?.so;/usr/local/openresty/luajit/lib/lua/5.1/?.so;/usr/local/lib/lua/5.1/loadall.so
+ENV GO_VERSION="1.13.12"
```
hutchic, line 12: can / should this be snagged from `.requirements` or no?

test/Dockerfile.test
gszr: This Dockerfile is used for running Kong tests, which contains an integration test for Go plugins.

test/Dockerfile.test
gszr: Don't think so, as this is only used as a test dependency at the moment

test/Dockerfile.test
hutchic: right good point. Are we going to eventually migrate this to `curl the go-pluginserver` release from github or in both testing and using `go-pluginserver` is a custom Kong docker image the route to proceed with?

gszr: @hutchic, as for the first question, we could do that, since we already tag releases and have the sources available in GH releases. But what we are doing here has the same effect of fetching the tag. Does that answer?

PR #308, opened 2020-07-06
fix(openresty-patches) add missing patch for FFI-based ngx.crc32 API

thibaultcha:
* add the patch enabling FFI-based hashing APIs in stream subsystem
  (prerequisite).
* add the subsequent patch enabling FFI-based ngx.crc32 API used by the
  dynamic upstream keepalive patch.


hishamhm:
Merged as part of #309.

PR #310, opened 2020-07-08
test(zlib) just a test commit to verify we can drop zlib-dev

hutchic:
Testing if it's safe to drop zlib-dev from Ubuntu https://github.com/Kong/docker-kong/tree/chore/drop-zlib-dev


hutchic:
turns out it's not safe

PR #319, opened 2020-08-25
Increase ARM64 lightuserdata mapping from 8 to 16 sectors

javierguerragiraldez:


openresty-patches/patches/1.15.8.3/LuaJIT-2.1-20190507_03-dyn-lightuserdata-map.patch
```diff
@@ -74,36 +75,35 @@ index 13b8230..05eed3a 100644
...
-@@ -830,6 +830,21 @@ typedef union GCobj {
+@@ -830,6 +830,20 @@ typedef union GCobj {
  #define setpriV(o, i)		(setitype((o), (i)))
  #endif
  
 +#ifdef LJ_LUV_SECT_SHIFT
-+#define LJ_LUV_SECT_MASK (7UL << LJ_LUV_SECT_SHIFT)
-+extern uint8_t luv_sect_map[8];
++#define LJ_LUV_SECT_MASK (((1UL << LJ_LUV_SECT_BITS) -1UL) << LJ_LUV_SECT_SHIFT)
++extern uint8_t luv_sect_map[1 << LJ_LUV_SECT_BITS];
```
kikito, line 87: If we declare `LJ_LUV_NUM_SECTS` in this file, we can use it in these calculations instead of `1 << LJ_LUV_SECT_BITS`, which would make them a bit more clear IMHO. I don't know how to do that with `UL`.

openresty-patches/patches/1.15.8.3/LuaJIT-2.1-20190507_03-dyn-lightuserdata-map.patch
```diff
@@ -113,26 +113,26 @@ index a0fba2a..1ab3097 100644
 +
 +static char checked_layout = 0;
 +
-+#define LJ_LUV_SECT_MASK (7UL << LJ_LUV_SECT_SHIFT)
-+uint8_t luv_sect_map[8] = {
++#define LJ_LUV_NUM_SECTS (1 << LJ_LUV_SECT_BITS)
```
kikito, line 116: Move this to `lj_arch.h` so it can be used in other places.

openresty-patches/patches/1.15.8.3/LuaJIT-2.1-20190507_03-dyn-lightuserdata-map.patch
```diff
@@ -182,7 +183,9 @@ index a0fba2a..1ab3097 100644
 +
 +  free(p);
 +
-+  luvmap_fold();
++  if (!luvmap_fold()) {
```
kikito, line 186: This (plus the replacement of `abort()` by `return NULL`) feels like an extra change, in addition to the change in size. Depending on how different this is from the previous behavior (which I have no way to tell) we might want to include a note in the PR about it, or even add a separate commit for it.

kikito (changes requested):

openresty-patches/patches/1.15.8.3/LuaJIT-2.1-20190507_03-dyn-lightuserdata-map.patch
javierguerragiraldez: done

openresty-patches/patches/1.15.8.3/LuaJIT-2.1-20190507_03-dyn-lightuserdata-map.patch
javierguerragiraldez: done

openresty-patches/patches/1.15.8.3/LuaJIT-2.1-20190507_03-dyn-lightuserdata-map.patch
javierguerragiraldez: the behaviour is the same: refuse to start if it can't find a suitable mapping.  the difference is that doing `abort()` results in a coredump, while now it just returns a non-success result.

I'd like to put a nice error message, but that implies either pulling `<stdio.h>` on this file (which currently only does internal-structs handling), or a much more invasive patch to add new failure modes.

kikito (approved):

PR #331, opened 2020-09-22
chore(cache) include kong*.rockspec to cache hash calculation

bungle:
### Summary

If it is not included in cache hash calculation, dependencies that are updated in .rockspec are leaked to other tests / builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ CACHE_BUSTER?=`date +%V`
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 TEST_SHA=$$(git log -1 --pretty=format:"%h" -- ${ROOT_DIR}/test/)${CACHE_BUSTER}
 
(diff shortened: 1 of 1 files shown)


hutchic (changes requested):
I don't believe this is nescessary.

`REQUIREMENTS_SHA` is used only by `DOCKER_OPENRESTY_SUFFIX` which is only used by `Dockerfile.openresty` which doesn't do anything with the `rockspec` file.

`Dockerfile.kong` is where we do `luarocks install` stuff whose cache should be covered by 
`DOCKER_KONG_SUFFIX=${BUILD_TOOLS_SHA}${OPENRESTY_PATCHES}${DEBUG}-${KONG_VERSION}-${KONG_SHA}-${CACHE_BUSTER}`
which includes
`KONG_SHA=$$(git --git-dir=$(KONG_SOURCE_LOCATION)/.git rev-parse --short HEAD)`

PR #339, opened 2020-10-01
feat(*) allow running kong as the 'kong' user

murillopaula:
Previously, Kong was, by default, running as root on non-docker systems,
which is not a security best practice. For docker environments, we take
steps to provide more security [2] (create a kong user by default, grant
the same /usr/local/kong/* files and directories ownership to that user
and to the root group) and these are similar the same steps that are being
employed here for non-docker environments. Post-installation fpm scripts
are being leveraged to create the kong user and change the permissions
of the kong package's contents to that user and to the root group. See
the fpm wiki [1] for details. See all the the files and directories
owned by the kong package here: [files.txt](https://github.com/Kong/kong-build-tools/files/5314643/files.txt)

Note that when `kong start` is run without changing the `$USER`, the
nginx master process will be run as the 'root' user and the nginx
workers processes will run as the 'nobody' user. That is the default
behavior in Kong because the nginx master process needs to run as 'root'
in order for nginx to execute certain actions (listen on privileged
ports such as port 80, for example).

In order to run kong as the 'kong' user, first switch to the kong user:

```
su kong
```

Then start `kong` normally:

```
kong start
```

This will make both the nginx master and worker processes run as the
'kong' user.

[1] https://github.com/jordansissel/fpm/wiki
[2] https://github.com/Kong/docker-kong/blob/2.1.3/rhel/Dockerfile#L36-L43


fpm-entrypoint.sh
```diff
@@ -18,6 +18,7 @@ elif [ "$RESTY_IMAGE_BASE" == "centos" ] || [ "$RESTY_IMAGE_BASE" == "rhel" ] ||
   fi
   if [ "$RESTY_IMAGE_BASE" == "amazonlinux" ]; then
     OUTPUT_FILE_SUFFIX=".aws"
+    FPM_PARAMS="$FPM_PARAMS -d /usr/sbin/useradd"
```
hutchic, line 21: this should move to lines 10 and 14 I think right (or somewhere after line 32 would work)?

We want to make `/usr/sbin/useradd` a dependency for everything but alpine correct?

murillopaula: `useradd` is not available by default on `amazonlinux` so that should be the only place where we need to install it, the other distros have it.

murillopaula:
Please revert or change these commits before merging:

- https://github.com/Kong/kong-build-tools/pull/339/commits/66abe497fdbbc414a128c1269cdbc1a9ec35d07a
- https://github.com/Kong/kong-build-tools/pull/339/commits/7c76f22158501f274c2790cebc908b7c818406f7

I'd suggest a squash and merge.

PR #363, opened 2020-11-23
fix(image) fix typo in label name

Tieske:
see https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

PR #364, opened 2020-11-23
fix(ci): correct the docker image label to match the spec

hutchic:
Adjust to match the spec https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys


hutchic:
cc @Tieske

Tieske:
see #363

PR #365, opened 2020-11-26
fix(tests) go path

gszr:
Explicitly set `GOPATH` and `GOROOT` to different directories; while the former is the Go development workspace, the latter carries the Go SDK; previously, errors like `cannot download, $GOPATH must not be set to $GOROOT` would occur when building in GOPATH mode.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+kong/servroot*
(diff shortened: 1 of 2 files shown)


javierguerragiraldez (approved):
LGTM

PR #375, opened 2020-12-18
chore(test) install vegeta for stress tests

locao:
Add the `vegeta` tool to the test environment.


kikito (approved):

kikito:
What do you think about installing vegeta as an extra step of `make dev` in the main Kong repo?

dndx:
@kikito I think we can leave this in the Docker container for now, installing too much stuff for `make dev` isn't really necessary because some user may only want to make simple changes and not all user needs to run the stress tests. In fact that's what setting up CI is for anyway.

dndx (approved):

PR #376, opened 2021-01-18
fix(tests) revert usage of build vars in rhel tests

gszr:
adapt to https://github.com/Kong/docker-kong/pull/447


gszr:
this is no longer needed

PR #377, opened 2021-01-26
tests(api) replace tavern tests with bash

gszr:


javierguerragiraldez (approved):
LGTM.

removing extra frameworks is always nice.
the use of `networks:` in the `-compose.yaml` is also a good thing, but i guess could make "manual" tests slightly more difficult.  OTOH, the simpler script chain will make it easier to use it properly, if well documented.

gszr:
> the use of networks: in the -compose.yaml is also a good thing, but i guess could make "manual" tests slightly more difficult. OTOH, the simpler script chain will make it easier to use it properly, if well documented.

It's a little easier to manually test now, as the admin and proxy ports are exposed. Previously, with the networking mode "host", it wasn't possible on Mac, due to how the Docker engine runs in a Linux VM (so "host" means the VM, not the VM's host :)

PR #379, opened 2021-02-10
Add geoip2 module

Elyahou:


fffonion:
Thanks for the PR @Elyahou ! I haven't compiled the module by myself yet, but it seems from the doc we will need to either compile libmaxminddb library or ensure the system has libmaxminddb headers installed. We used to compile all dependencies like pcre and openssl by ourself but I'm not sure if we should do the same (related to my comment at https://github.com/Kong/kong/issues/6789#issuecomment-776925656).

Elyahou:
@fffonion You are right, libmaxminddb is a required dependency, so it need to be installed beforehand.

and to answers to your comments [here](https://github.com/Kong/kong/issues/6789#issuecomment-776925656), I didn't tried to use it as a dynamic module but only as a regular module, but it seems to be interesting.

In my PR, I let the possibility to not include the geoip module, if we want to make it the default, I can change it to be disabled by default and enabled only if selected. But then we will need to decide whether to enable it by default for docker images etc...

fffonion:
@Elyahou let's disable it by default so it won't stop user from building it, if they don't have libmaxminddb headers installed.

@hutchic I would also appreciate your thoughts on this one, (also the origin issue at https://github.com/Kong/kong/issues/6789#issuecomment-776925656)

gszr:
Hey @Elyahou. It'd be interesting to run a few tests running this module as a dynamically loaded module, and seeing how it meshes with Kong. Could you run such tests and update this thread? IMO, this route would be preferable; as @fffonion pointed out, this wouldn't require us adding a new dependency, requiring less maintenance burden. Please do let us know if you see any blockers for running it as a dynamic module, and let's keep the discussion open.

Elyahou:
@gszr As far I understand, using the geoip2 module as a dynamic module will still require the libmaxminddb dependency...

As I see it, dynamic should be compiled against the correct nginx version in the `kong-ngx-build` tool and be placed to some directory to be then be loaded at runtime using nginx conf file.

I am not sure what is the pros of using dynamic modules in place of static modules for kong.

codemug:
@Elyahou has there been any progress on this work?

PR #395, opened 2021-04-29
chore(release) replace bintray with pulp

gszr:
This PR accomplishes two main goals:
- Replaces Bintray with our internal distribution server in the release script.
- Removes integration tests stages, similarly to what has been done in kong/kong, following the ADR https://github.com/Kong/core-team/pull/26/files.


hutchic (approved):
I think you'll need to update `PULP_HOST` as it currently defaults to staging.

Although .... we could modify the `when`'s in this Jenkinsfile to do a test using our staging pulp instance

kikito (approved):

release-kong.sh
```diff
@@ -1,230 +1,125 @@
...
 set -eo pipefail
 
+
 CWD=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+BUILD_DIR="$CWD/output"
+
+
+PULP_USERNAME="$PULP_USR"
+PULP_PASSWORD="$PULP_PSW"
+PULP_DOCKER_IMAGE="kong/release-script"
```
hutchic, line 11: I kind of wish we had a `kong/release-scripts:1` so this can be pinned but only to the major. I suppose it's fine it's an image we own so we should know to not break backwards compatibility

hutchic (approved):
LGTM

Goodbye bintray

PR #410, opened 2021-10-19
add BoringSSL option

javierguerragiraldez:


hutchic:
THIS IS EXCITING 🎉

openresty-build-tools/kong-ngx-build
```diff
@@ -158,44 +168,74 @@ main() {
...
+    OPENSSL_INSTALL=$(canon_path $OPENSSL_DESTDIR/$OPENSSL_PREFIX)
+
+    if version_lt $OPENSSL_VER 1.1; then
+      # unconditionally disable module in older core since they are never tested
+      KONG_NGINX_MODULE=0
+    fi
+  fi
+
+  if [ -n "$BORINGSSL_VER" ]; then
+    OPENSSL_DOWNLOAD=$DOWNLOAD_CACHE/boringssl-$BORINGSSL_VER
```
kikito, line 190: Why is `OPENSSL_DOWNLOAD` being "reused" here for boring_ssl, instead of using a separate `BORINGSSL_DOWNLOAD` variable?
```suggestion
    BORINGSSL_DOWNLOAD=$DOWNLOAD_CACHE/boringssl-$BORINGSSL_VER
```

openresty-build-tools/kong-ngx-build
```diff
@@ -158,44 +168,74 @@ main() {
...
+#    OPENSSL_DESTDIR=${OPENSSL_DESTDIR:-/}
+#    OPENSSL_PREFIX=${OPENSSL_PREFIX:-$PREFIX/openssl}
+#    OPENSSL_INSTALL=$(canon_path $OPENSSL_DESTDIR/$OPENSSL_PREFIX)
```
kikito, line 193: These don't seem needed, so I suggest removing:
```suggestion
```

openresty-build-tools/kong-ngx-build
```diff
@@ -158,44 +168,74 @@ main() {
...
+  fi
+
+  if version_lt $NGINX_CORE_VER 1.13.6; then
```
kikito, line 196: The previous version of this if had an extra condition:

```
if version_lt $NGINX_CORE_VER 1.13.6 || version_lt $OPENSSL_VER 1.1; then
```

Is the `version_lt $OPENSSL_VER 1.1` part not needed any more? I expected it to "reappear" on the `if [ -n "$OPENSSL_VER" ];` conditional, below.

openresty-build-tools/kong-ngx-build
```diff
@@ -158,44 +168,74 @@ main() {
...
+        fi
+        tar -xzf openssl-$OPENSSL_VER.tar.gz
       fi
-      set -e
+    fi
+
+    if [ -n "$BORINGSSL_VER" ]; then
+      # Boring SSL
+
+      if [[ ! -d "$OPENSSL_DOWNLOAD" ]]; then
```
kikito, line 228:
```suggestion
      if [[ ! -d "$BORINGSSL_DOWNLOAD" ]]; then
```

openresty-build-tools/kong-ngx-build
```diff
@@ -380,56 +420,88 @@ main() {
...
-    succ "OpenSSL $OPENSSL_VER has been built successfully!"
+    else
+      succ "OpenSSL $OPENSSL_VER has been built successfully (cached)!"
+    fi
+  fi
 
-  else
-    succ "OpenSSL $OPENSSL_VER has been built successfully (cached)!"
+  if [ -n "$BORINGSSL_VER" ]; then
+    if [ ! -f "$OPENSSL_DOWNLOAD/.openssl/lib/libcrypto.so" ]; then
```
kikito, line 478:
```suggestion
    if [ ! -f "$BORINGSSL_DOWNLOAD/.openssl/lib/libcrypto.so" ]; then
```

openresty-build-tools/kong-ngx-build
```diff
@@ -380,56 +420,88 @@ main() {
...
+      notice "Building BoringSSL..."
+
+      pushd $OPENSSL_DOWNLOAD
```
kikito, line 481:
```suggestion
      pushd $BORINGSSL_DOWNLOAD
```

openresty-build-tools/kong-ngx-build
javierguerragiraldez: the idea was that boringssl is finally in the same place as openssl would be, so the nginx build phase is mostly unchanged.    but to be honest, i don't think it's actually that similar.  i'll try using separate variables and see if it improves readability.

openresty-build-tools/kong-ngx-build
javierguerragiraldez: boringssl is forked from 1.0.something.  i'll move this condition to the openssl part

javierguerragiraldez: the full version of this condition is a few lines above

kikito (approved):

openresty-build-tools/kong-ngx-build
```diff
@@ -380,56 +417,88 @@ main() {
...
+
+      pushd $BORINGSSL_DOWNLOAD
+
+        mkdir -p build
+        pushd build
+          CC=/usr/bin/clang \
+          CXX=/usr/bin/clang++ \
+          CPP=/usr/bin/clang-cpp \
+          LD=/usr/bin/ld.lld \
+          cmake -GNinja -DBUILD_SHARED_LIBS=1 -DFIPS=1  ..
```
fffonion, line 486: for a general purpose boringssl build this should be fine, but if i understand correctly for fips 
we will need to follow the exact steps as performed in the security policy
(https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp3678.pdf section 12.1), that uses specific version
of build chain. actually we can reuse much of that from envoy (https://github.com/envoyproxy/envoy/blob/a12869fa9e9add4301a700978d5489e6a0cc0526/bazel/external/boringssl_fips.genrule_cmd).
Might need to downgrade the c++ headers as well, the fips validated module doesn't compile
happy with most recent libstdc++ headers (https://github.com/fffonion/lua-resty-openssl/blob/master/.github/workflows/tests.yml#L156 could be helpful)

javierguerragiraldez: yes, full validation requires those steps and verification, but the 2.7 scope isn't FIPS-validation, just to run with the library in FIPS mode.  the exact version can be specified in the `.requirements` file; perhaps the C++ headers version should be there too?

fffonion: >  just to run with the library in FIPS mode.

in that case this feels okay to me.

> the exact version can be specified in the .requirements file;

Yeah locking the versions in .requirements is indeed a good approach for future 👍

PR #417, opened 2021-10-30
add support for building against preinstalled OpenSSL

saju:
Adds support for OpenResty to be built against a previously
built/customized build of OpenSSL by introducing a new optional
argument --openssl-dir which can pointed to an existing
OPENSSL_INSTALL location.

### DIFF
--- a/openresty-build-tools/kong-ngx-build
+++ b/openresty-build-tools/kong-ngx-build
@@ -7,6 +7,7 @@ PREFIX=
 DOWNLOAD_CACHE=work
 OPENRESTY_VER=
 OPENSSL_VER=
+OPENSSL_DIR=
 LUAROCKS_VER=
 PCRE_VER=
 FORCE=0
@@ -59,6 +60,10 @@ main() {
         ;;
       --openssl_sha)
         OPENSSL_SHA=$2
+        shift 2
+        ;;
+      --openssl-dir)
+        OPENSSL_DIR=$2
         shift 2
         ;;
       --luarocks)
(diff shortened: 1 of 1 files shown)


openresty-build-tools/kong-ngx-build
```diff
@@ -660,9 +686,13 @@ show_usage() {
   echo "Required arguments:"
   echo "  -p, --prefix <prefix>              Location where components should be installed."
   echo "      --openresty <openresty_ver>    Version of OpenResty to build, such as 1.13.6.2."
-  echo "      --openssl <openssl_ver>        Version of OpenSSL to build, such as 1.1.1c."
   echo ""
   echo "Optional arguments:"
+  echo "      --openssl <openssl_ver>        Version of OpenSSL to build, such as 1.1.1c."
+  echo ""
+  echo "      --openssl-dir <openssl_path>   Path to an existing OpenSSL installation."
+  echo "                                     Atleast one of --openssl or --openssl-dir must be supplied."
```
guanlan, line 694: nitpicking: At least

saju: changed, squashed & re-pushed

hutchic:
if memory serves me correctly -- we compile / built openssl from source because we historically had to patch it for the previous version of Kong mesh built upon the Kong gateway (vs the current Kong mesh which is build upon kuma).

I think it's worth revisiting if we need to compile / build openssl from scratch at all and if it couldn't be set as a dependency

hutchic (approved):

javierguerragiraldez:
> how does this related to our BoringSSL work?

In my tests, i put borinssl options as mutually exclusive with openssl.   it would be nice to just make boringssl a "user supplied" version of openssl, but there are some other compiler and `configure` options to be set differently, so it's easier to manage separately.

in short, this doesn't affect much because the boringssl option replaces it completely.

PR #420, opened 2021-11-08
chore(enterprise): adjustments for Kong enterprise build process compatibility

hutchic:
various adjustments to move us towards compatibility with Kong Enterprise build process.

- need cmake in base images
- need yq in base images
- need to compensate for the logrotate file naming


dockerfiles/Dockerfile.deb
```diff
@@ -28,4 +28,16 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update \
...
-        valgrind
+        valgrind \
+        wget
+
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.15.7/cmake-3.15.7.tar.gz \
+    && tar -xzvf cmake-3.15.7.tar.gz \ 
+    && cd cmake-3.15.7 \
+    && ./bootstrap \
+    && make -j 2 \
+    && make install
```
mikefero, line 39: Same here; WDYT about allowing the version to be specified, switching to `curl` (or ensure no certificate checks), and cleanup the files?

```suggestion
ARG CMAKE_VERSION=3.15.7
RUN curl -sSL https://github.com/Kitware/CMake/releases/download/v$CMAKE_VERSION/cmake-$CMAKE_VERSION.tar.gz  -o /tmp/cmake-$CMAKE_VERSION.tar.gz \
    && tar -xzvf /tmp/cmake-$CMAKE_VERSION.tar.gz -C /tmp \ 
    && pushd /tmp/cmake-$CMAKE_VERSION \
    && ./bootstrap \
    && make -j 2 \
    && make install \
    && popd \
    && rm -rf /tmp/cmake-$CMAKE_VERSION.tar.gz /tmp/cmake-$CMAKE_VERSION
```

or

```suggestion
ARG CMAKE_VERSION=3.15.7
RUN wget --no-check-certificate -O /tmp/cmake-$CMAKE_VERSION.tar.gz https://github.com/Kitware/CMake/releases/download/v$CMAKE_VERSION/cmake-$CMAKE_VERSION.tar.gz \
    && tar -xzvf /tmp/cmake-$CMAKE_VERSION.tar.gz -C /tmp \ 
    && pushd /tmp/cmake-$CMAKE_VERSION \
    && ./bootstrap \
    && make -j 2 \
    && make install \
    && popd \
    && rm -rf /tmp/cmake-$CMAKE_VERSION.tar.gz /tmp/cmake-$CMAKE_VERSION
```


You might also consider changing the `make -j 2` to use the actual number of cores available or remove the `-j 2` all together; unless of course we are guaranteed that 2 cores are always available:

#### without `nproc`

This is very ugly and unpredictable, but gives you the actual number of physical cores

```
grep ^cpu\\scores /proc/cpuinfo | uniq |  awk '{print $4}'
```

#### with `nproc`

This is a lot simpler and won't break if the `/proc/cpuinfo` format changes. The only negative here is that it will return all cores if hyperthreading is enabled.

**Note**: Ensure `coreutils` is installed

```
nproc
```

README.md
```diff
@@ -42,7 +42,7 @@ The most common ones are the following:
 
 ```
 RESTY_IMAGE_BASE=ubuntu|centos|rhel|debian|alpine|amazonlinux
-RESTY_IMAGE_TAG=xenial|bionic|focal|6|7|8|stretch|buster|latest|latest
+RESTY_IMAGE_TAG=18.04|20.04|6|7|8|9|10|11|latest|latest
```
mikefero, line 45: why not 16.04 as well?

```suggestion
RESTY_IMAGE_TAG=16.04|18.04|20.04|6|7|8|9|10|11|latest|latest
```

test/build_container.sh
```diff
@@ -34,6 +35,11 @@ if [ "$RESTY_IMAGE_TAG" == "stretch" ]; then
   sed -i 's/unzip git/unzip git gdebi/g' docker-kong/ubuntu/Dockerfile
 fi
 
+if [ "$RESTY_IMAGE_BASE" == "debian"] && [ "$RESTY_IMAGE_TAG" == "9" ]; then
+  sed -i 's/apt install --yes /gdebi -n /g' docker-kong/ubuntu/Dockerfile
+  sed -i 's/unzip git/unzip git gdebi/g' docker-kong/ubuntu/Dockerfile
+fi
+
```
mikefero, line 42: Shouldn't this just have been an update the above `stretch` conditional?

mikefero (commented):
Won't changing all the `RESTY_IMAGE_TAG` from names to numbers cause confusions with all of the current tags?

hutchic:
> Won't changing all the RESTY_IMAGE_TAG from names to numbers cause confusions with all of the current tags?
There will be a few unecessary builds because of a cache miss but overall they'll be fine.

This repository expires cache every week as is anyways so it'll be a small blip

README.md
hutchic: mainly because it's old and eol.

test/build_container.sh
hutchic: yeah I think that's what I meant to do ... but failed. I dunno will fix lol

mikefero: nope backwards compatible makes complete sense; I was referring to exactly what you were "supposing"

mikefero: > if [ "$RESTY_IMAGE_TAG" == "stretch" ] || [ "$RESTY_IMAGE_TAG" == "9" ]; then

We should probably future proof this as there is a  RHEL 9 on the horizon ([a beta was just released](https://www.redhat.com/en/blog/whats-new-rhel-90-beta)) and a `RESTY_IMAGE_TAG=9` there would apply to the conditional above. For complete backwards compatibility we could do this and remove the addition in this PR:

```
if [ "$RESTY_IMAGE_TAG" == "stretch" ] ||
  { [ "$RESTY_IMAGE_BASE" == "debian" ] && [ "$RESTY_IMAGE_TAG" == "9" ]; } then
  sed -i 's/apt install --yes /gdebi -n /g' docker-kong/ubuntu/Dockerfile
  sed -i 's/unzip git/unzip git gdebi/g' docker-kong/ubuntu/Dockerfile
fi
```

hutchic: This hax is only necessary for earlier than stretch so when stretch is EOL is should go away

README.md
mikefero: Ubuntu 16.04 has only reached end of standard support not EOL; that doesn't occur until [April 2026](https://wiki.ubuntu.com/Releases).

test/build_container.sh
mikefero: But we also need `jessie` (8) which neither are ETLS yet:

- Debian 8 (jessie) ELTS is 2022/06/30
- Debian 9 (stretch) ETLS hasn't been defined yet

see see https://wiki.debian.org/LTS/Extended for more details.

Jenkinsfile
```diff
@@ -144,7 +144,6 @@ pipeline {
                     steps {
                         sh 'mkdir -p /home/ubuntu/bin/'
                         sh 'git clone --single-branch --branch ${KONG_SOURCE} https://github.com/Kong/kong.git ${KONG_SOURCE_LOCATION}'
-                        sh 'export RESTY_IMAGE_TAG=8 && make package-kong && make test && make cleanup'
```
mikefero: You can add the following to the overall docker file to fix the OpenSSL issues with `curl` and `wget`:

```
RUN sed -i'' 's/^mozilla\/DST_Root_CA_X3.crt$/!mozilla\/DST_Root_CA_X3.crt/g' /etc/ca-certificates.conf
RUN update-ca-certificates
```

See https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/ for more details.

Makefile
```diff
@@ -143,6 +143,10 @@ else ifeq ($(BUILDX),true)
 endif
 
 build-base:
+	docker pull centos:8
+	docker tag centos:8 centos:7
+	docker pull debian:9
+	docker tag debian:9 debian:8
```
mikefero, line 149: This is a little scary and seems misleading; since the toolchain versions are different in CentOS 7<->8 and Debian 8<->9 we are more than likely going to create incompatibility when installing packages built on newer versions of a distro and installed on older distros. Has this been tested with package installs?

hutchic: yup the CI tests it.

As discussed at stnadup openssl is the reason we can't use a single debian binary for all debian distributions but centos 7/8 are compatible as well as debian 8/9. This gives us the advantage of using non-EOL base distributions so there's less hax's while still being able to build / backport Kong to them

.ci/setup_ci.sh
```diff
@@ -2,7 +2,7 @@
 
 set -x
 
-if [ "$RESTY_IMAGE_TAG" != "xenial" ] && [ "$RESTY_IMAGE_BASE" != "alpine" ]; then
+if [ "$RESTY_IMAGE_TAG" != "xenial" ] && [ "$RESTY_IMAGE_TAG" != "16.04" ] && [ "$RESTY_IMAGE_BASE" != "alpine" ]; then
```
curiositycasualty, line 5: I'm having a hard time interpreting the intent here.
This seems to be meant to be a no-op on xenial or alpine, but it's unclear why.

A comment here would be appreciated.

Jenkinsfile
```diff
@@ -72,9 +73,10 @@ pipeline {
                         }
                     }
                     steps {
+                        sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
                         sh 'git clone --single-branch --branch ${KONG_SOURCE} https://github.com/Kong/kong.git ${KONG_SOURCE_LOCATION}'
                         sh 'export RESTY_IMAGE_BASE=src RESTY_IMAGE_TAG=src PACKAGE_TYPE=src && make package-kong && make test && make cleanup'
-                        sh 'export RESTY_IMAGE_BASE=alpine RESTY_IMAGE_TAG=3 PACKAGE_TYPE=apk CACHE=false UPDATE_CACHE=true DOCKER_MACHINE_ARM64_NAME="jenkins-kong-"`cat /proc/sys/kernel/random/uuid` && make package-kong && make test && make cleanup'
+                        sh 'export RESTY_IMAGE_BASE=alpine RESTY_IMAGE_TAG=3.10 PACKAGE_TYPE=apk CACHE=false UPDATE_CACHE=true DOCKER_MACHINE_ARM64_NAME="jenkins-kong-"`cat /proc/sys/kernel/random/uuid` && make package-kong && make test && make cleanup'
```
curiositycasualty, line 79: A comment here explaining why the explicit major/minor version pin is happening here would be appreciated.

Makefile
```diff
@@ -50,6 +50,8 @@ ifndef AWS_ACCESS_KEY
 	BUILDX=false
 else ifeq ($(RESTY_IMAGE_TAG),xenial)
 	BUILDX=true
+else ifeq ($(RESTY_IMAGE_TAG),16.04)
+	BUILDX=true
```
curiositycasualty, line 54: Is this because we're only producing the arm builds on ubunti?

curiositycasualty (approved):
Just a couple minor recommendations, otherwise 👍

Jenkinsfile
mikefero: I think we should create a ticket to fix this in Kong Enterprise; I think @RobSerafini did a POC during the SSL certificate failures we were hitting back in September.

mikefero (approved):
Looks good :crossed_fingers:

.ci/setup_ci.sh
hutchic: The setup_ci.sh file isn't a great name (anymore).

The file sets up ci systems such that they're capable of building / releasing arm assets which we only do for alpine and xenial

PR #428, opened 2021-12-16
feat(dl): add --download-extract-only flag

curiositycasualty:
This adds a flag to the `openresty` build script entitled: `--download-extract-only`: the purpose of which is to seed the download cache with the various source archives this script collects, and exiting early to avoid also compiling those sources.
Subsequent runs of this script, assuming `--work`/`$DOWNLOAD_CACHE` is the same, will avoid re-download/extracting said files via the numerous `test -d` `if/then/fi` checks that already exist here.

Ideally, this would have been solely a `--download-only` flag but the current structure of the script doesn't lend itself to early termination/return (as it place entirely within a single `main()` function).

### DIFF
--- a/openresty-build-tools/kong-ngx-build
+++ b/openresty-build-tools/kong-ngx-build
@@ -5,6 +5,7 @@ set -e
 SCRIPT_PATH=
 PREFIX=
 DOWNLOAD_CACHE=work
+DOWNLOAD_ONLY=
 OPENRESTY_VER=
 OPENSSL_VER=
 BORINGSSL_VER=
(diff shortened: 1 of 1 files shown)


javierguerragiraldez (approved):

openresty-build-tools/kong-ngx-build
```diff
@@ -321,6 +337,11 @@ main() {
     fi
   popd
 
+  if [ ! -z "$DOWNLOAD_ONLY" ]; then
+    notice "Exiting early since --download-extract-only was specified"
+    exit 0
+  fi
+
```
hutchic, line 344: should the early exit happen after applying patches?

curiositycasualty: I did debate on where to put this escape hatch.

The patching isn't gated by the existence of the extracted download directories the same way the download & extraction is. `patch` is pretty forgiving when you ask it to re-apply a patch that is already in place too. So I'm unconcerned about pushing the patching to subsequent runs.

PR #433, opened 2022-01-11
fix(openresty) cjson.empty_array is nil when run on Apple M1/aarch64

chronolaw:
When I am testing on Apple M1 + VMware + ubuntu 22.04,
I found cjson.empty_array can not work as expect,

I debug the cjson's code, then find the reason,
but we can not wait the openresty's release,
so I did this patch, If openresty's bundle fix this bug later,
then we must remove this patch.


mayocream:
Tests passed with this patch on my Mac M1, great!

bungle:
©chronolaw, all our current patches are sorted. Can you please rename:
```
lua-cjson-2.1.0.8-empty_array.patch
```
to
```
lua-cjson-2.1.0.8_01-empty_array.patch
```

Otherwise this is great, and ready to be merged. Thank you.

bungle (approved):

PR #439, opened 2022-01-26
feat(dockerfiles): move our dockerfiles refs: #FT-2144

hutchic:
The base images have moved https://github.com/Kong/kong-build-tools-base-images

~Requires https://github.com/Kong/docker-kong/pull/538~
~Requires https://github.com/Kong/docker-kong/pull/541~


Makefile
```diff
@@ -48,15 +48,15 @@ DOCKER_MACHINE_ARM64_NAME?=docker-machine-arm64-${USER}
...
-endif
+#ifndef AWS_ACCESS_KEY
+#	BUILDX=false
+#else ifeq ($(RESTY_IMAGE_TAG),xenial)
+#	BUILDX=true
+#else ifeq ($(RESTY_IMAGE_TAG),16.04)
+#	BUILDX=true
+#else ifeq ($(RESTY_IMAGE_BASE),alpine)
+#	BUILDX=true
+#endif
```
hutchic, line 59: todo -- comment back in

dockerfiles/Dockerfile.openresty
```diff
@@ -1,9 +1,8 @@
-ARG RESTY_IMAGE_BASE="ubuntu"
-ARG RESTY_IMAGE_TAG="20.04"
 ARG DOCKER_BASE_SUFFIX
 ARG DOCKER_REPOSITORY
+ARG PACKAGE_TYPE
 
-FROM ${DOCKER_REPOSITORY}:${RESTY_IMAGE_BASE}-${RESTY_IMAGE_TAG}-${DOCKER_BASE_SUFFIX}
+FROM hutchic/temp:$PACKAGE_TYPE
```
hutchic, line 5: todo -- set to somewhere official

hutchic:
Waiting on https://github.com/Kong/docker-kong/pull/538 before this will go green again

test/build_container.sh
```diff
@@ -26,6 +26,7 @@ elif [ "$RESTY_IMAGE_BASE" == "alpine" ]; then
   cp output/*.${ARCHITECTURE}.apk.tar.gz docker-kong/alpine/kong.tar.gz
   BUILD_DIR="alpine"
 elif [ "$RESTY_IMAGE_BASE" == "centos" ] || [ "$RESTY_IMAGE_BASE" == "amazonlinux" ]; then
```
kikito, line 28: It seems we still use `$RESTY_IMAGE_BASE` instead of `$PACKAGE_TYPE` is still being used in the tests, isn't that problematic?

kikito: It seems problematic: I was going to ask why removing the `sed` line was needed in (we remove them for for testing amazonlinux assets, says the commit message).

hutchic: when testing an asset we've produced we use git://kong/docker-kong as a test harness, however the Dockerfile's there in some cases are close to a correct harness less some differences. Because the docker-kong Dockerfile's are used for our official docker image builds there's some restrictions that don't allow them to be as flexible as we require.

All of that is a long way of saying I HATE the spots we employ `sed` but haven't found a suitable alternative.

The other spot we will still require RESTY_IMAGE_BASE and RESTY_IMAGE_TAG is in the fpm script as it needs to capture differences that occur between OS versions such as a dependent package was renamed etc

kikito: `sed` is convenient _in the moment_: it gets the job done, but it often requires you to have the whole "tree of dependencies" inside your head. `sed` gives trouble _in the future_, when someone who doesn't have the "tree of dependencies" fresh in their heads (can be future you) tries to make sense of it.

> haven't found a suitable alternative.

It might be that you are looking for an alternative that is convenient both _in the moment_ and _in the future_.

I think we should consider using a solution that makes things more maintainable - even if it makes things more complex in the short term. Usually this means putting the "dependencies" out of our heads and making them explicit in the code.

A recent example I can think of: In docker-kong, we use `sed` inside our update.sh script to get the download urls of each distro, directly from their Dockerfiles. The problem with that is that that's implicit knowledge. If you look at the Ubuntu Dockerfile, you cannot know that `update.sh` depends on it - that the "source of truth for the ubuntu download url" is the Dockerfile itself. So you might [change the url calculation in a way that breaks the sed in update.sh, and not notice anything](https://github.com/Kong/docker-kong/pull/537/commits/34e715e8eb060650f6f28954272c454b7aa11981). This implicit knowledge makes things difficult to maintain.

The alternative,is making those dependencies more explicit. We could introduce a file of "shared globals". Which could contain stuff like:

```
# globals.sh

def ubuntu_download_url() {
  KONG_VERSION=$1
  UBUNTU_CODENAME=$2
  echo "https://download.konghq.com/gateway-${KONG_VERSION%%.*}.x-
ubuntu-${UBUNTU_CODENAME}/pool/all/k/kong/kong_${KONG_VERSION}_amd64.deb"
}

def centos_download_url(){
  ...
}
```
(This is just an example, the exact way in which we share stuff might be different).

Once we have that we could use`globals.sh` in both `update.sh` and when setting parameters for the `Dockerfile`. This would make dependencies more explicit: the inclusion in `update.sh` would be direct. The Dockerfile would need a `DOWNLOAD_URL` ENV, which would be provided by a `make_dockerfile.sh` script, that would include `globals.sh`.

But that is way more complicated than using a simple `sed` inside update.sh. But that is the tradeoff: In order to avoid problems in the future we have to make sacrifices in the present.

Makefile
```diff
@@ -40,7 +40,7 @@ KONG_NGINX_MODULE ?= `grep KONG_NGINX_MODULE $(KONG_SOURCE_LOCATION)/.requiremen
 RESTY_LMDB ?= `grep RESTY_LMDB $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 LIBYAML_VERSION ?= `grep LIBYAML_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 OPENRESTY_PATCHES ?= 1
-DOCKER_KONG_VERSION = 'master'
+DOCKER_KONG_VERSION = 'fix/centos-yum-clean'
```
hutchic, line 43: pending https://github.com/Kong/docker-kong/pull/541
```suggestion
DOCKER_KONG_VERSION = 'master'
```

kikito (approved):
lgtm. One question though: what's the reason for moving the base images to a different repo (https://github.com/Kong/kong-build-tools-base-images)? Wouldn't it have been easier to keep them on this one?

hutchic:
@kikito yes and no it was mostly done for better developer experience

The images are shared between kong-build-tools and kong-distributions (at least until kong-distributions is retired). 

Our base images seldom change but we wanted it to be a better developer experience. So the kong-build-tools-base-images repository has pre-commit including pre-commit CI. It has automated releases based on semmantic messaging to both github tag as well as docker hub that include automated release notes.

While all of that could have been done in this repository it would have been "crowded". For example we would want the pre-commit CI as well as the release automation to only run on the three base Docker images and vice-versa we wouldn't want the kong-build-tools jenkins integration to kick off for changes related to the base docker images or their CI configuration

PR #442, opened 2022-02-08
chore(boringssl) use gcc - FT-2342

gszr:
Use gcc instead of clang.


gszr:
note: need to squash + merge

javierguerragiraldez (approved):
great!

PR #445, opened 2022-03-08
fix(group): adjust Kong group ownership to be root owned

hutchic:
related: https://github.com/Kong/docker-kong/pull/546

This changes ownership of Kong installed files from the current `kong:kong` to `kong:root` such that kong isn't able to rewrite it's own code

~could this be considered a breaking change?~

hold merging this until 3.0

### DIFF
--- a/after-install.sh
+++ b/after-install.sh
@@ -17,7 +17,7 @@ create_user() {
   FILES="${FILES} /usr/local/share/lua/"
 
   for FILE in ${FILES}; do
-    chown -R kong:kong ${FILE}
+    chown -R kong:root ${FILE}
     chmod -R g=u ${FILE}
   done
 
--- a/after-install.sh
+++ b/after-install.sh
@@ -17,8 +17,8 @@ create_user() {
   FILES="${FILES} /usr/local/share/lua/"
 
   for FILE in ${FILES}; do
-    chown -R kong:root ${FILE}
-    chmod -R g=u ${FILE}
+    chown -R root:kong ${FILE}
+    chmod -R g=u,g-w ${FILE}
   done
 
   return 0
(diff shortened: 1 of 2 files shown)


after-install.sh
```diff
@@ -17,7 +17,7 @@ create_user() {
   FILES="${FILES} /usr/local/share/lua/"
 
   for FILE in ${FILES}; do
-    chown -R kong:kong ${FILE}
+    chown -R kong:root ${FILE}
     chmod -R g=u ${FILE}
```
dannysauer, line 21: Given that the group permissions are set to mirror the user permissions, it's conceivable that something is using the group perms to access the file contents.  So changing the group ownership from `kong` to `root` would break those things.

More importantly, this should probably be root:kong, followed by a chmod which removes group write.  That way, anything accessing as group kong can still read and execute as before, but can no longer write.  The kong user should be a member of the kong group, so anything formerly running as `kong` will still be able to read and execute by virtue of being in the kong group.
```suggestion
    chown -R root:kong ${FILE}
    chmod -R g=u,g-w ${FILE}
```

fpm-entrypoint.sh
```diff
@@ -3,10 +3,12 @@
 set -o errexit
 
 cd /tmp/build
+chgrp root -R /tmp/build/*
+chmod -R g=u /tmp/build/*
```
dannysauer, line 7: These two probably oughtta be changed to match the `after-install.sh` ones. :)

dannysauer: Are there any dot files at the top level of those directories the `/*` would miss?  This pattern should also include dotfiles at the top level containing at least two characters after the dot. :)
```suggestion
chown -R root:kong /tmp/build/{,.??}*
chmod -R g=u,g-w /tmp/build/{,.??}*
```

PR #451, opened 2022-03-24
feat(slimages): migrate to the slim docker-kong Dockerfiles refs: #FT-2378

hutchic:


Makefile
```diff
@@ -40,7 +40,7 @@ KONG_NGINX_MODULE ?= `grep KONG_NGINX_MODULE $(KONG_SOURCE_LOCATION)/.requiremen
 RESTY_LMDB ?= `grep RESTY_LMDB $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 LIBYAML_VERSION ?= `grep LIBYAML_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 OPENRESTY_PATCHES ?= 1
-DOCKER_KONG_VERSION = 'master'
+DOCKER_KONG_VERSION = 'feat/slimages'
```
hutchic, line 43: before merging
```suggestion
DOCKER_KONG_VERSION = 'master'
```

aboudreault (approved):

tyler-ball (approved):

test/tests/01-package/run.sh
```diff
@@ -97,13 +97,6 @@ docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "ldd /
 docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "ldd /usr/local/openresty/bin/openresty | grep -q /usr/local/openresty/luajit/lib/libluajit-5.1.so.2"
 docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "/usr/local/openresty/bin/openresty -V 2>&1 | grep /work/pcre-${RESTY_PCRE_VERSION}"
 
-# luarocks
-docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "luarocks --version | grep -q ${RESTY_LUAROCKS_VERSION}"
-docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "luarocks config | grep -q /usr/local/openresty/luajit/bin/luajit"
-docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "luarocks config | grep -q /usr/local/openresty/luajit/include/luajit-2.1"
-docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "luarocks config | grep -q /usr/local/openresty/luajit/lib"
-docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "luarocks install version"
-
```
kikito, line 110: I would like to understand the amount of Breaking in `BREAKING CHANGE luarocks in Kong 3.0 will not be operational`.

If the `luarocks` executable is literally not in the image any more, that will mean that:
* people can not use our images as a base and then add a `luarocks install my-custom-kong-plugin`. I don't think we list that as a way to add custom plugins (in our instructions we tell people to put their files in the right place, manually)
* people can not use luarocks to install dependencies for their plugins (e.g the kafka plugin needing the kafka luarock).
* It will also mean that we will no longer be able to run `make dev` inside a kong image, if got this right.

Will luarocks be available as opt-in feature, perhaps? Something along the lines of `install_luarocks.sh && make dev`

hutchic: This is discussed in https://konghq.atlassian.net/browse/ENGEN-449 but the TLDR is the `luarocks` binary is still available but it will be hobbled either by chown'd to `root:root` and/or chmod to `0`

> Installing custom plugins & dependencies

We will adapt [customize](https://github.com/Kong/docker-kong/tree/master/customize) for this purpose and update relevant documentation for folks that aren't using customize.

> no longer able to run `make dev`

Good that means it worked :)

Official production images must not contain or be capable of doing development out of the box. We will pick a sensible default (likely `ubuntu:focal`) for a "fat" development image that lands on the unofficial release channels (ie `kong/kong:dev`) but with 3.0 we want to draw a very hard line between "production image" and "development image"

kikito: Or is it root-only? `sudo chmod +x luarocks && sudo chown $USER luarocks && make dev`

curiositycasualty: I mean, because there's an operating system in a docker container, there's pretty much no way to exclude a "root" user without a lot of butchering-- which means users will always have an option similar to what you described, of selecting the user with which a container is run-- and that user can always be "root".

kikito: Understood. Thanks for explaining.

kikito (approved):

PR #452, opened 2022-03-24
feat(kong-ee): adapt kong-build-tools to be more extendable

hutchic:


Jenkinsfile
```diff
@@ -50,147 +26,145 @@ pipeline {
...
+                        PACKAGE_TYPE = "rpm"
                     }
                     steps {
                         sh 'echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin || true'
-                        sh 'git clone --single-branch --branch ${KONG_SOURCE} https://github.com/Kong/kong.git ${KONG_SOURCE_LOCATION}'
-                        sh 'export RESTY_IMAGE_BASE=src RESTY_IMAGE_TAG=src PACKAGE_TYPE=src && make package-kong && make test && make cleanup'
-                        sh 'export RESTY_IMAGE_BASE=alpine RESTY_IMAGE_TAG=3.10 PACKAGE_TYPE=apk CACHE=false UPDATE_CACHE=true DOCKER_MACHINE_ARM64_NAME="jenkins-kong-"`cat /proc/sys/kernel/random/uuid` && make package-kong && make test && make cleanup'
+                        sh 'while /bin/bash -c "ps aux | grep [a]pt-get"; do sleep 5; done'
+                        sh 'curl https://raw.githubusercontent.com/Kong/kong/master/scripts/setup-ci.sh | bash'
+                        sh 'git clone --recursive --single-branch --branch ${KONG_SOURCE} https://github.com/Kong/kong-ee.git ${KONG_SOURCE_LOCATION}'
```
hutchic, line 46: todo: when the kong-ee submodules are gone this is no longer necessary

Jenkinsfile
```diff
@@ -4,7 +4,7 @@ pipeline {
         cron(env.BRANCH_NAME == 'master' ? '@weekly' : '')
     }
     environment {
-        KONG_SOURCE = "master"
+        KONG_SOURCE = "feat/branch-by-abstraction"
```
hutchic, line 7: after kong-ee merges set this to master

Makefile
```diff
@@ -62,17 +78,17 @@ endif
...
+	DOCKER_COMMAND?=docker buildx build --progress=$(DOCKER_BUILD_PROGRESS) --push --platform="linux/amd64,linux/arm64"
 endif
 
 # Set this to unique value to bust the cache
 CACHE_BUSTER?=0
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 TEST_SHA=$$(git log -1 --pretty=format:"%h" -- ${ROOT_DIR}/test/)${CACHE_BUSTER}
 
-REQUIREMENTS_SHA=$$(md5sum $(KONG_SOURCE_LOCATION)/.requirements | cut -d' ' -f 1)
+REQUIREMENTS_SHA=$$(find kong/distribution -type f -print0 | sort -z | xargs -0 sha256sum | sha256sum  | cut -d' ' -f 1)
```
flrgh, line 91: Q: What file(s) is this supposed to checksum?

build-kong.sh
```diff
@@ -10,9 +10,9 @@ rocks_trees = {
 }
 " > $ROCKS_CONFIG
 
+[ -d "/usr/local/share/lua/" ] && cp -R /usr/local/share/lua/ /tmp/build/usr/local/share/
```
flrgh, line 13: Could be wrong, but I _think_ this will cause errexit (`set -e`) to bail out and fail if `/usr/local/share/lua` does not exist. Is that the intended behavior? If not I think this must be wrapped in an `if`

```suggestion
if [[ -d /usr/local/share/lua ]]; then
  cp -R /usr/local/share/lua/ /tmp/build/usr/local/share/ 
fi
```

openresty-build-tools/kong-ngx-build
```diff
@@ -232,6 +232,7 @@ main() {
           notice "Downloaded: $(sha256sum "openssl-$OPENSSL_VER.tar.gz")"
         fi
         tar -xzf openssl-$OPENSSL_VER.tar.gz
+        ln -s openssl-${OPENSSL_VER} openssl
```
flrgh, line 235: Nice! Looks like a good way to let the later stages of the build stop caring about `$DEPENDENCY_VERSION`.

Is there a reason to symlink instead of rename? I don't really have an underlying preference--mostly just curious.

build-kong.sh
```diff
@@ -89,7 +55,18 @@ popd
 cp /kong/COPYRIGHT /tmp/build/usr/local/kong/
 cp /kong/bin/kong /tmp/build/usr/local/bin/kong
 sed -i 's/resty/\/usr\/local\/openresty\/bin\/resty/' /tmp/build/usr/local/bin/kong
+sed -i 's/\/tmp\/build//g' /tmp/build/usr/local/bin/openapi2kong || true
```
flrgh, line 58: This isn't a real suggestion because I know there's a pattern here that nobody wants to deviate from, but... it'd be awesome to see these `sed` invocations using a different pattern delimiter so that `/` doesn't need to be escaped. I like `|` or `#`

```suggestion
sed -i 's#/tmp/build##g' /tmp/build/usr/local/bin/openapi2kong || true
```

Makefile
hutchic: The entire distribution directory. So in the case of kong it will only contain the `.requirements` file but in the case of EE all of https://github.com/Kong/kong-ee/tree/feat/branch-by-abstraction/distribution such that our cache key invalidates when relevant parts of the build process are changed

openresty-build-tools/kong-ngx-build
hutchic: it gives verbosity and convenience:
verbosity `ls` will tell you the specific openssl version convenience `cd openssl` works.

We also require the versioned directory for backwards compatibility reasons :(

build-kong.sh
hutchic: huh yeah in hindsight that's an excellent suggestion. Per the PR description we (Isa and I) are avoiding refactors as much as we can in this series of PR's but for sure. I'd like to lint the entire repository :laughing:

test/tests/03-enterprise-licensing-tests/run.sh
```diff
@@ -0,0 +1,51 @@
...
+msg_test "Create a service"
+assert_response "-d name=testservice -d url=http://mockbin $KONG_ADMIN_URI/services" "201"
+
+msg_test  "List services"
+assert_response "$KONG_ADMIN_URI/services" "200"
+
+msg_test "Create a route"
+assert_response "-d name=testroute -d paths=/ $KONG_ADMIN_URI/services/testservice/routes" "201"
+
+msg_test "List services"
```
yzyyi, line 19: shall we test the "List routes" here

build-kong.sh
```diff
@@ -89,7 +57,18 @@ popd
...
 grep -l -I -r '\/tmp\/build' /tmp/build/
 sed -i 's/\/tmp\/build//' `grep -l -I -r '\/tmp\/build' /tmp/build/`
 
 chown -R 1000:1000 /tmp/build/*
+
+# bytecompile if script is present (as with ee)
+if test -f "/distribution/post-bytecompile.sh"; then
```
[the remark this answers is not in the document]
hutchic: There's the same amount of test protection against such an error as already currently exists -- which is to say none ;)

If we have enough concern I think it a worthy inclusion as a follow up PR though I personally suspect we have larger more pressing test blindspots

test/tests/03-enterprise-licensing-tests/run.sh
hutchic: added list routes in 4b4c703a51a4bcf64bcd8ced3173d44533b5c8a5

flrgh (approved):
:+1: this looks good to me, though I think getting another +1 from a gateway dev with a longer tenure than me before merging would be wise.

yzyyi (approved):

Jenkinsfile
```diff
@@ -13,32 +13,19 @@ pipeline {
         DEBUG = 0
     }
     options {
+        retry(1)
```
hutchic, line 16: to future historians: added a retry not because our code is flaky but the internet in general has a habit of failing exactly when the end to end smoke tests require it to not

PR #454, opened 2022-04-06
feat(openresty) bundle Kong/lua-resty-websocket

flrgh:
This adds [Kong's fork of lua-resty-websocket](https://github.com/Kong/lua-resty-websocket]) to kong-ngx-build.

Main differences between our fork and the current lua-resty-websocket release (v0.08) are:

* mTLS support for `client:connect()` - this has been [merged](https://github.com/openresty/lua-resty-websocket/pull/72) but is not yet released
* other `client:connect()`  enhancements - features needed by Kong that we'll attempt to upstream as best we can ([example](https://github.com/openresty/lua-resty-websocket/pull/76))

Full diff:
https://github.com/openresty/lua-resty-websocket/compare/v0.08...Kong:master

This affects CE and EE as both rely on the mTLS functionality. Up until now this has been achieved by way of [patch file](https://github.com/Kong/kong-build-tools/blob/886101503c3bc96577ada04124bd812260701016/openresty-patches/patches/1.19.9.1/lua-resty-websocket-0.08_01-client-mtls.patch).

---

Feedback wanted:

* This fork and our existing lua-resty-websocket patch files must be mutually-exclusive. An earlier iteration of this code simply removed the relevant patch file(s). This is nice and straightforward but not backwards compatible. The current iteration is fully opt-in and attempts to skip `lua-resty-websocket` patch files when using our fork, so it's backwards-compatible but.... eh, feels messy. How much concern is backwards compatibility: any/some/none?
* I've tried to be as defensive as possible here, but any ideas for how to do additional testing/sanity-checking?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ KONG_GMP_VERSION ?= `grep KONG_GMP_VERSION $(KONG_SOURCE_LOCATION)/.requirements
 KONG_NETTLE_VERSION ?= `grep KONG_NETTLE_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 KONG_NGINX_MODULE ?= `grep KONG_NGINX_MODULE $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 RESTY_LMDB ?= `grep RESTY_LMDB $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
+RESTY_WEBSOCKET ?= `grep RESTY_WEBSOCKET $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 LIBYAML_VERSION ?= `grep LIBYAML_VERSION $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 OPENRESTY_PATCHES ?= 1
 DOCKER_KONG_VERSION = 'master'
@@ -109,6 +110,7 @@ debug:
 	@echo ${DEBUG}
 	@echo ${KONG_NGINX_MODULE}
 	@echo ${RESTY_LMDB}
+	@echo ${RESTY_WEBSOCKET}
 
 setup-ci: setup-build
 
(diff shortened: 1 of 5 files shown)


hutchic:
is it feasible / useful to add a test?

curiositycasualty (approved):
This all looks correct and in keeping with the current spirit of the script.

At first I was going to add a nit about optimistically writing these portions of the script in anticipation of reverting to the upstream repo, but that doesn't quite make sense.

flrgh:
> is it feasible / useful to add a test?

@hutchic I'm not sure. The existing tests same to revolve around validating the state of a container/image that has been built for us, whereas all my test cases would revolve around running a few one-off builds before checking the final state (or validating that the build fails in the way we expect it to):

1. build without `--resty-websocket`, check the FS to ensure the correct version is present (for some known OpenResty version)
2. build with `--resty-websocket x.y.z`, check the FS to ensure the correct version is present
3. build with `--resty-websocket x.y.z` for a version that doesn't exist, assert build failure
4. build with `--resty-websocket x.y.z` for a version of OpenResty that doesn't already bundle it (if one exists) and assert build failure

Seems like a lot more shell scripting would be needed to properly test those cases, unfortunately. The good news is that this library is depended upon in CE and EE, so a failure to install the desired version of the library while bubble up in an integration test failure over there.

Would it be worth adding a simple FS stat check [here](https://github.com/Kong/kong-build-tools/blob/548f3b8a9bd077967b5ea68386f55582378eb9f3/test/tests/01-package/run.sh#L93)? We won't be able to validate the version of the files since it won't be known, but so long as we always install OpenResty at the same path, we can validate their existence.

hutchic:
> @flrgh Would it be worth adding a simple FS stat check [here](https://github.com/Kong/kong-build-tools/blob/548f3b8a9bd077967b5ea68386f55582378eb9f3/test/tests/01-package/run.sh#L93)?

Though naive I think that would suffice and be a worthy inclusion. The earlier we receive feedback that something is amiss the better

flrgh:
@hutchic okay I think this is ready for another review

Validated the correct behavior by exercising both code paths (default and opt-in w/ RESTY_WEBSOCKET={{some-known-tag}}) in different commits, letting CI run, and then confirming the expected [version](https://github.com/Kong/kong-build-tools/blob/417b28c70f88c71bc05d65375c7cd1fbfc7a52ee/test/tests/01-package/run.sh#L102) shows up in the logs.

hutchic (approved):

PR #457, opened 2022-04-11
fix(useradd) make sure kong:kong exist and files ownership is correct

outsinre:
If the group kong exists before installation, "useradd -U" would not create the kong user.
The file ownership update would also fail.

Firstly create the kong group; then create the kong user on demand.


after-install.sh
```diff
@@ -9,7 +10,9 @@ create_user() {
   FILES="${FILES} /usr/local/bin/lua2json"
   FILES="${FILES} /usr/local/bin/luarocks"
   FILES="${FILES} /usr/local/bin/luarocks-admin"
+  FILES="${FILES} /usr/local/bin/openapi2kong"
   FILES="${FILES} /usr/local/etc/luarocks/"
+  FILES="${FILES} /usr/local/etc/passwdqc/"
```
hutchic, line 15: these two addition are enterprise only are they not?

currently KBT build kong
kong-distributions builds kong enterprise

that changes in 3.0 (hopefully) with https://github.com/Kong/kong-build-tools/pull/455

hutchic (changes requested):
We have a desire to go the other direction with this in 3.0 https://github.com/Kong/kong-build-tools/pull/445

At a glance this PR looks fine to go in until then if we desire though (pending the one change)

after-install.sh
outsinre: @hutchic , yep, currently it is. I added this two lines to make "kong-distributions" (EE) and "kong-build-tools" (CE) synced.

According to @gszr, we will deprecate "kong-distributions" in favor of "kong-build-tools". So, along this PR, I added the extra two lines.

A similar PR to is created for the enterprise version now: https://github.com/Kong/kong-distributions/pull/773

hutchic (approved):
approved with a note that 3.0 moves us from `kong:kong` to `kong:root`

gszr:
> approved with a note that 3.0 moves us from `kong:kong` to `kong:root`

👋 curious what the reason is for the change?

hutchic:
@gszr the desire that the application not be capable of rewriting itself while it runs. It's a highly unlikely attack vector but relatively easy to protect against

curiositycasualty:
I'm hoping https://github.com/Kong/kong-build-tools/pull/463/files sets this PR up for success.

hutchic:
Going to approve this and the kong-build-tools PR's bug label them appropriately so we don't merge until 3.0

hutchic (approved):
approved pending 3.0 release

outsinre:
> @gszr the desire that the application not be capable of rewriting itself while it runs. It's a highly unlikely attack vector but relatively easy to protect against

@hutchic, could you share similar attack cases? Any links are welcome.

dannysauer:
@outsinre 
> @hutchic, could you share similar attack cases? Any links are welcome.

To start with, it should be `root:kong`, with the group having read (and execute, in the case of a directory or executable) permission.  The files should be owned by a user which is not the same user the process is running as.  That allows files to be modified by an administrator who can assume the identity of that user, while still allowing the process (which should be a member of the specified group) to be able to read.  In general, this is an example of applying "the principle of least privilege", where each actor has only the minimum set of privileges necessary to complete tasks, and no more.  ACLs would really be better, but those are somewhat poorly understood in general, and not universally supported.

For a moderately concrete example, I'll note that this is a very common strategy with web servers and other services that watch their config file for changes.  Consider a web service run by the web server.  A coding mistake allows a remote user to change the storage path of some data which is normally written to a specific folder dedicated to the app. If only the path can be changed, it's possible that the attacker could change the written path to the web service's configuration file.  In the "best" case, this allows a remote attacker to break the server's configuration by writing whatever garbage over the config file.  In a worse case, the attacker controls the data and maybe causes this web server to start exposing /etc/passwd or user homedirs or whatever else.

Notably, services which read their config file at startup and then don't look again often have their config set as `root:root` (or some other combination which is outside of the process's privilege set).  The approach there is to start the process as root.  The process reads its config and binds to privileged ports (and whatever other things require increased privileges), then drops privileges by either switching to a lower-permission user or using the relevant capability bounding set syscalls.

Basically, giving processes more privileges than they need opens up the risk of processes doing more things than they should do, and that risk increases as more entities have access to interact with the process (like, say, a gateway exposed to the Internet). :)

after-install.sh
```diff
@@ -9,15 +10,17 @@ create_user() {
...
   FILES="${FILES} /usr/local/kong/"
   FILES="${FILES} /usr/local/lib/lua/"
   FILES="${FILES} /usr/local/lib/luarocks/"
   FILES="${FILES} /usr/local/openresty/"
   FILES="${FILES} /usr/local/share/lua/"
 
   for FILE in ${FILES}; do
-    chown -R kong:kong ${FILE}
+    chown -R kong:root ${FILE}
     chmod -R g=u ${FILE}
```
dannysauer, line 24: Implementing the control described in the comment at https://github.com/Kong/kong-build-tools/pull/457#issuecomment-1125126539, the chmod should really be `chmod -R g=u,g-w` unless the group actually needs write access.

Related: why is this in a `for` loop?  Just run like `chown $FILES` and let the shell's whitespace splitting work it out; that's already what's happening in the for loop anyway, and there's no sense fork/exec'ing chown & chmod 15 times. ;)

dannysauer: One point is that we shouldn't be making new folders at the filesystem root to begin with.  But perhaps that's a separate problem. :D

dannysauer: Option 2 is the higher security approach.  Using root:root with other access gives access to all processes on the entire machine.  Using root:kong with the world access bits removed limits access to only members of the kong group.

The reason nginx provides world access to its config is that nginx is a general-purpose web server.  Its use cases include serving user home directories and being run as a proxy by other processes on the system.  So it is a service which may need to be run by any user on the system while including its global config.  Kong gateway is a different use case - it is intended to be run by only the Kong administrator, and then any non-administrative users who interact with it are expected to only use the API.  So, the only users who need to access its files directly are Kong administrators and the process itself.

dannysauer: FWIW: I checked the Fedora, Debian, arch, and SUSE (ok, I was a SUSE release coordinator, so I actually just remembered that one) packaging guides earlier today.  They all say root:root/0755 for binaries unless increased security is necessary (a suid or sgid binary, for example).  None have a specific guideline for configuration, so I think we're generally free to do what makes most sense to us.

outsinre: > One point is that we shouldn't be making new folders at the filesystem root to begin with. But perhaps that's a separate problem. :D

Maybe, we should let Kong creates runtime data folder dynamically under `/run/` or `/var/run`. It's the de facto. Agree that, this is another topic.

outsinre: > FWIW: I checked the Fedora, Debian, arch, and SUSE (ok, I was a SUSE release coordinator, so I actually just remembered that one) packaging guides earlier today. They all say root:root/0755 for binaries unless increased security is necessary (a suid or sgid binary, for example). None have a specific guideline for configuration, so I think we're generally free to do what makes most sense to us.

Ha, really appreciate the time to do the research. Leaving it as default (`root:root`) is preferable, as that will make our Kong package as much compatible with different Linux distributions as possible.

I will file a commit to this PR.

outsinre:
This is partially reverted in https://github.com/Kong/kong-build-tools/pull/468.

outsinre:
The `kong:kong` to `kong:root` is reverted to keep consistent with https://github.com/Kong/kong-distributions/pull/783.

New PR should be filed to care about the file permission issue.

after-install.sh
dndx: In that case I recommend changing the:

```
chown -R kong:kong ${FILE}
chmod -R g=u ${FILE}
```

below to:

```
chown -f -R kong:kong ${FILE}
chmod -f -R g=u ${FILE}
```

so to avoid potential errors during OSS install.

PR #460, opened 2022-04-21
feat(openresty) bundle Kong/lua-resty-events

chronolaw:
This adds [Kong/lua-resty-events](https://github.com/Kong/lua-resty-events) to kong-ngx-build.


hutchic (approved):
approved with a note that adding a test would be nice if possible

fffonion (approved):

PR #462, opened 2022-04-29
feat(LuaJIT) change the LuaJIT version format to something like `LuaJIT 2.1-20220411`

ADD-SP:
This is one solution to fix `FT-2807`. Another solution is https://github.com/Kong/kong-build-tools/pull/466 .

```
$ resty -e "print(jit.version)"
LuaJIT 2.1.0-20210510
```

### DIFF
--- a/openresty-build-tools/kong-ngx-build
+++ b/openresty-build-tools/kong-ngx-build
@@ -585,6 +585,15 @@ main() {
 
     pushd $OPENRESTY_DOWNLOAD
       if [ ! -f Makefile ]; then
+
+        pushd bundle
+          lj_dir=$(ls -d LuaJIT*)
+          lj_release_date=$(echo ${lj_dir} | sed -e 's/LuaJIT-[[:digit:]]\+.[[:digit:]]\+-\([[:digit:]]\+\)/\1/')
+          lj_version_tag="LuaJIT\ 2.1.0-${lj_release_date}"
+        popd
+
+        notice "Building LuaJIT with version string $lj_version_tag"
+
         OPENRESTY_OPTS=(
           "--prefix=$OPENRESTY_PREFIX"
           "--with-pcre-jit"
@@ -593,6 +602,7 @@ main() {
           "--with-http_stub_status_module"
           "--with-http_v2_module"
           "--without-http_encrypted_session_module"
+          "--with-luajit-xcflags='-DLUAJIT_VERSION=\\\"${lj_version_tag}\\\"'"
           "-j$NPROC"
         )
 
--- a/openresty-patches/patches/1.19.9.1/LuaJIT-2.1-20210510_03_patch_macro_luajit_version.patch
+++ b/openresty-patches/patches/1.19.9.1/LuaJIT-2.1-20210510_03_patch_macro_luajit_version.patch
@@ -1,0 +1,27 @@
+From f53c8fa441f4233b9a3f19fcd870207fe8795456 Mon Sep 17 00:00:00 2001
+From: Qi <add_sp@outlook.com>
+Date: Wed, 25 May 2022 18:35:08 +0800
+Subject: [PATCH] Patch macro `LUAJIT_VERSION`
+
+---
+ src/luajit.h | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/LuaJIT-2.1-20210510/src/luajit.h b/LuaJIT-2.1-20210510/src/luajit.h
+index a4d33001..e35f4e7e 100644
+--- a/LuaJIT-2.1-20210510/src/luajit.h
++++ b/LuaJIT-2.1-20210510/src/luajit.h
+@@ -32,7 +32,9 @@
+
+ #define OPENRESTY_LUAJIT
+
++#ifndef LUAJIT_VERSION
+ #define LUAJIT_VERSION		"LuaJIT 2.1.0-beta3"
++#endif
+ #define LUAJIT_VERSION_NUM	20100  /* Version 2.1.0 = 02.01.00. */
+ #define LUAJIT_VERSION_SYM	luaJIT_version_2_1_0_beta3
+ #define LUAJIT_COPYRIGHT	"Copyright (C) 2005-2022 Mike Pall"
+--
+2.34.1
+
+
--- a/openresty-patches/patches/1.21.4.1/LuaJIT-2.1-20210510_01_patch_macro_luajit_version.patch
+++ b/openresty-patches/patches/1.21.4.1/LuaJIT-2.1-20210510_01_patch_macro_luajit_version.patch
@@ -1,0 +1,27 @@
+From f53c8fa441f4233b9a3f19fcd870207fe8795456 Mon Sep 17 00:00:00 2001
+From: Qi <add_sp@outlook.com>
+Date: Wed, 25 May 2022 18:35:08 +0800
+Subject: [PATCH] Patch macro `LUAJIT_VERSION`
+
+---
+ src/luajit.h | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/LuaJIT-2.1-20220411/src/luajit.h b/LuaJIT-2.1-20220411/src/luajit.h
+index a4d33001..e35f4e7e 100644
+--- a/LuaJIT-2.1-20220411/src/luajit.h
++++ b/LuaJIT-2.1-20220411/src/luajit.h
+@@ -32,7 +32,9 @@
+
+ #define OPENRESTY_LUAJIT
+
++#ifndef LUAJIT_VERSION
+ #define LUAJIT_VERSION		"LuaJIT 2.1.0-beta3"
++#endif
+ #define LUAJIT_VERSION_NUM	20100  /* Version 2.1.0 = 02.01.00. */
+ #define LUAJIT_VERSION_SYM	luaJIT_version_2_1_0_beta3
+ #define LUAJIT_COPYRIGHT	"Copyright (C) 2005-2022 Mike Pall"
+--
+2.34.1
+
+
--- a/test/tests/01-package/run.sh
+++ b/test/tests/01-package/run.sh
@@ -97,6 +97,8 @@ docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "ldd /
 docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "ldd /usr/local/openresty/bin/openresty | grep -q /usr/local/kong/lib/libcrypto.so.1.1"
 docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "ldd /usr/local/openresty/bin/openresty | grep -q /usr/local/openresty/luajit/lib/libluajit-5.1.so.2"
 docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "/usr/local/openresty/bin/openresty -V 2>&1 | grep /work/pcre-${RESTY_PCRE_VERSION}"
+docker run ${USE_TTY} --user=root --rm ${KONG_TEST_IMAGE_NAME} /bin/sh -c "/usr/local/openresty/bin/resty -e 'print(jit.version)' | grep -q 'LuaJIT[[:space:]][[:digit:]]\+.[[:digit:]]\+.[[:digit:]]\+-[[:digit:]]\{8\}'"
+
 
 # lua-resty-websocket library (sourced from OpenResty or Kong/lua-resty-websocket)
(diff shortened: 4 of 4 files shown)


chronolaw (approved):
LGTM.

hutchic (approved):
I don't _think_ this is a breaking change right?

flrgh (commented):
Is it possible to add a basic test for this? Maybe validate the output of `luajit -v` somewhere in [here](https://github.com/Kong/kong-build-tools/blob/master/test/tests/01-package/run.sh)?

openresty-build-tools/kong-ngx-build
```diff
@@ -275,6 +275,13 @@ main() {
           notice "Downloaded: $(sha256sum "openresty-$OPENRESTY_VER.tar.gz")"
         fi
         tar -xzf openresty-$OPENRESTY_VER.tar.gz
+
```
flrgh, line 278: nit: some trailing whitespace here needs cleanup

openresty-build-tools/kong-ngx-build
```diff
@@ -275,6 +275,13 @@ main() {
           notice "Downloaded: $(sha256sum "openresty-$OPENRESTY_VER.tar.gz")"
         fi
         tar -xzf openresty-$OPENRESTY_VER.tar.gz
+        
+        # change the luajit version text
+        pushd openresty-$OPENRESTY_VER/bundle
+        lj_dir=$(ls -d LuaJIT*)
+        lj_update_time=$(echo ${lj_dir} | sed -e 's/LuaJIT-[[:digit:]]\+.[[:digit:]]\+-\([[:digit:]]\+\)/\1/')
+        sed -i "s/#define[[:space:]]\+LUAJIT_VERSION[[:space:]]\+\"LuaJIT \([[:digit:]]\+.[[:digit:]]\+\).[[:digit:]]\+.\+\"/#define LUAJIT_VERSION		\"LuaJIT v\1-${lj_update_time}\"/" ${lj_dir}/src/luajit.h
+        popd
```
flrgh, line 284: nit/style: lines between `pushd`/`popd` should be indented 2 spaces to match the style of the rest of this script

```suggestion
        pushd openresty-$OPENRESTY_VER/bundle
          lj_dir=$(ls -d LuaJIT*)
          lj_update_time=$(echo ${lj_dir} | sed -e 's/LuaJIT-[[:digit:]]\+.[[:digit:]]\+-\([[:digit:]]\+\)/\1/')
          sed -i "s/#define[[:space:]]\+LUAJIT_VERSION[[:space:]]\+\"LuaJIT \([[:digit:]]\+.[[:digit:]]\+\).[[:digit:]]\+.\+\"/#define LUAJIT_VERSION		\"LuaJIT v\1-${lj_update_time}\"/" ${lj_dir}/src/luajit.h
        popd
```

flrgh: Some other bash/style notes here (feel free to disregard though)...

---

```bash
lj_update_time=$(echo ${lj_dir} | sed -e 's/LuaJIT-[[:digit:]]\+.[[:digit:]]\+-\([[:digit:]]\+\)/\1/')
```

This can be expressed more succinctly (and without  by using bash string functions:

```bash
lj_update_time=${lj_dir##*-}
```

---

```bash
sed -i "s/#define[[:space:]]\+LUAJIT_VERSION[[:space:]]\+\"LuaJIT \([[:digit:]]\+.[[:digit:]]\+\).[[:digit:]]\+.\+\"/#define LUAJIT_VERSION		\"LuaJIT v\1-${lj_update_time}\"/" ${lj_dir}/src/luajit.h
```

Maybe consider breaking this onto multiple lines and using the extended regex support in sed to make this pattern more readable with less escaping:

> -E, -r, --regexp-extended
>
>    use extended regular expressions in the script (for portability use POSIX -E).

(note: I didn't test this)

```bash
sed -i -r \
  -e "s/#define +LUAJIT_VERSION +\"LuaJIT ([0-9]+[.][0-9]+)[.][0-9]+.+\"/#define LUAJIT_VERSION		\"LuaJIT v\1-${lj_update_time}\"/" \
  ${lj_dir}/src/luajit.h
```

flrgh (commented):
I left some comments about bash-isms, but I'm actually changing my opinion a bit about this now...

I think this would be better as a static patch file instead of dynamic patching w/ `sed` at build time:

1. LuaJIT version is known for each OpenResty release and does not need to be discovered
2. We already have [LuaJIT patches in the repo](https://github.com/Kong/kong-build-tools/blob/master/openresty-patches/patches/1.19.9.1/LuaJIT-2.1-20210510_02-arm64-fix-pcall-error-case.patch) so this would follow an existing pattern
3. Future maintainers won't need to deal with reading/writing `sed` expressions--only creating a new patch file

Yes this means we need a new patch file for each release, but OpenResty releases are very infrequent. Test coverage can be added so that CI fails if a maintainer adds a new OpenResty version without a proper LuaJIT patch.

hutchic:
@flrgh I'm ok with either approach

> Future maintainers won't need to deal with reading/writing sed expressions--only creating a new patch file

Does seem appealing :thinking:

chronolaw:
> @flrgh I'm ok with either approach
> 
> > Future maintainers won't need to deal with reading/writing sed expressions--only creating a new patch file
> 
> Does seem appealing 🤔

Yes, I think it is a good idea. The patch may be easier to maintain than the shell script.

dndx:
@flrgh Static patch is not a good idea. This change is only cosmetic, even if it breaks, nothing will caught on fire. What we want is to automate the process as much as possible because I do not feel anyone will be updating these patch files for each release.

The frequency of new tags is surprisingly high, see: https://github.com/openresty/luajit2/tags. That would be a burden to carry forward to update the patch manually whenever we bump OpenResty.

flrgh:
> This change is only cosmetic, even if it breaks, nothing will caught on fire.

I disagree.

With a static patch file, the most probable breakage results in a patch file that doesn't apply cleanly, which will fail fast and produce a straightforward error.

With dynamic patching, breakage could easily introduce invalid syntax into luajit.h, resulting in the build failing with a compile-time error. This is a much more difficult issue to debug, especially for new maintainers who might be wholly unaware that we dynamically altered the file with `sed` at some point during the build.

> The frequency of new tags is surprisingly high, see: https://github.com/openresty/luajit2/tags. That would be a burden to carry forward to update the patch manually whenever we bump OpenResty.

If the luajit tag only changes when we bump OpenResty, then the frequency of new tags (for us) is actually very low.

> a static patch that requires human intervention every time (and remember doing so)

Is this not the workflow for all of our OpenResty patches, more-or-less (copy $old_release_patch_dir => $new_release_patch_dir, edit patch files as needed)? I don't see how this is a great burden.

dndx:
@flrgh Relying on human to do updates to patch file is error prone and tedious, and is something we try to avoid as much as possible. Not to mention currently we do not have patch file that require such special attention for every upgrade.

I do not know what's the difference between "With a static patch file, the most probable breakage results in a patch file that doesn't apply cleanly, which will fail fast and produce a straightforward error." and "With dynamic patching, breakage could easily introduce invalid syntax into luajit.h, resulting in the build failing with a compile-time error. This is a much more difficult issue to debug".

I recommend giving it a try and see if the error message in the later case is actually "harder to debug" than the first one. Considering how stable the LuaJIT release is nowadays, I would be surprised if this patch method would even breaks in the foreseeable future.

flrgh:
@dndx
Nobody else called it a bad idea or seemed to require so much additional convincing, so I am confident my point has been made and needs no further justification. I don't see the use in continuing to argue the merits of `sed` vs `patch` here. If you are confident your opinion is more correct than mine and the other maintainers who gave a :+1: to my suggestion, then feel free to merge this PR.

fffonion:
How about doing a patch like
```diff
diff --git a/src/luajit.h b/src/luajit.h
index a4d33001..e35f4e7e 100644
--- a/src/luajit.h
+++ b/src/luajit.h
@@ -32,7 +32,9 @@

 #define OPENRESTY_LUAJIT

+#ifndef LUAJIT_VERSION
 #define LUAJIT_VERSION         "LuaJIT 2.1.0-beta3"
+#endif
 #define LUAJIT_VERSION_NUM     20100  /* Version 2.1.0 = 02.01.00. */
 #define LUAJIT_VERSION_SYM     luaJIT_version_2_1_0_beta3
 #define LUAJIT_COPYRIGHT       "Copyright (C) 2005-2022 Mike Pall"
```
Then when building you can do `./configure --with-luajit-xcflags='-DLUAJIT_VERSION=\\"abaaba\\"'.

fffonion:
Alternatively we want to do patch-free, consider using `-D_LUAJIT_H` to bypass the luajit.h, then include it in your own header, then undef `LUAJIT_VERSION` and define it again; or the less hacky way:
```shell
mv src/luajit.h src/luajit_old.h
cat >> src/luajit.h << EOF
#ifndef _LUAJIT_H

#include "luajit_old.h"

#undef LUAJIT_VERSION
#define LUAJIT_VERSION         "LuaJIT 2.1.0-beta3-a-new-version"
#endif
EOF

**Comment by dndx** [comment]
@fffonion I think our goal here is to minimize the work and maintaineence burden for such a trivial change. So the `--with-luajit-xcflags='-DLUAJIT_VERSION=\"abaaba\"'` approach works fine for me as well. Still, the actual `abaaba` part should be generated automatically and not maintained by humans.

@ADD-SP Could you update the PR and maybe `kong-ngx-build` as well to incorporate this method?

**Comment by ADD-SP** [comment]
> @fffonion I think our goal here is to minimize the work and maintaineence burden for such a trivial change. So the `--with-luajit-xcflags='-DLUAJIT_VERSION=\"abaaba\"'` approach works fine for me as well. Still, the actual `abaaba` part should be generated automatically and not maintained by humans.
> 
> @ADD-SP Could you update the PR and maybe `kong-ngx-build` as well to incorporate this method?

done

File: openresty-build-tools/kong-ngx-build [hunk 3]
```diff
@@ -276,6 +276,25 @@ main() {
...
+          lj_dir=$(ls -d LuaJIT*)
+          lj_version_with_time=$(echo ${lj_dir} | sed -e 's/LuaJIT-\([[:digit:]]\+.[[:digit:]]\+-[[:digit:]]\+\)/\1/')
+          luajit_h_content=$(cat ${lj_dir}/src/luajit.h)
+          # mv ${lj_dir}/src/luajit.h ${lj_dir}/src/luajit_origin.h
+          cat > ${lj_dir}/src/luajit.h << EOF
+${luajit_h_content}
+
+#ifndef _LUAJIT_H_EX
+#define _LUAJIT_H_EX
+#ifdef _KONG_LUAJIT_VERSION_WITH_TIME
```
fffonion, line 290: let's use the value from macro to directly pass version

chronolaw:
Don't forget to add a patch in 1.21.4.1 when this PR is merged.

openresty-build-tools/kong-ngx-build
```diff
@@ -588,6 +594,7 @@ main() {
           "--with-http_stub_status_module"
           "--with-http_v2_module"
           "--without-http_encrypted_session_module"
+          "--with-luajit-xcflags='-DLUAJIT_VERSION=\\\"v${lj_version_with_time}\\\"'"
```
dndx, line 597: Is the `v` needed here? Will it make `LUAJIT_VERSION` `vLuaJIT-2.1-20210510` instead?

openresty-build-tools/kong-ngx-build
```diff
@@ -580,6 +580,12 @@ main() {
 
     pushd $OPENRESTY_DOWNLOAD
       if [ ! -f Makefile ]; then
+
+        pushd bundle
+          lj_dir=$(ls -d LuaJIT*)
+          lj_version_with_time=$(echo ${lj_dir} | sed -e 's/LuaJIT-\([[:digit:]]\+.[[:digit:]]\+-[[:digit:]]\+\)/\1/')
```
dndx, line 586:
```suggestion
          lj_version_tag=$(echo ${lj_dir} | sed -e 's/LuaJIT-\([[:digit:]]\+.[[:digit:]]\+-[[:digit:]]\+\)/\1/')
```

It is not "time" that we are including here.

dndx (changes requested):

dndx (approved):

openresty-build-tools/kong-ngx-build
```diff
@@ -580,6 +580,16 @@ main() {
 
     pushd $OPENRESTY_DOWNLOAD
       if [ ! -f Makefile ]; then
+
+        pushd bundle
+          lj_dir=$(ls -d LuaJIT*)
+          lj_update_time=$(echo ${lj_dir} | sed -e 's/LuaJIT-[[:digit:]]\+.[[:digit:]]\+-\([[:digit:]]\+\)/\1/')
+          lj_raw_version=$(cat ${lj_dir}/src/luajit.h | grep "#define LUAJIT_VERSION[[:space:]]" | sed -e 's/#define[[:space:]]\+LUAJIT_VERSION[[:space:]]\+\"LuaJIT \([[:digit:]]\+.[[:digit:]]\+.[[:digit:]]\+\).\+\"/\1/')
```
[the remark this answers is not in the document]
dndx: Also can we change the name to `lj_release_date` instead? It is not a time.

dndx (approved):

PR #466, opened 2022-05-03
feat(LuaJIT) change the LuaJIT version format to something like `LuaJIT v2.1-20220411`

ADD-SP:
This is one solution to fix `FT-2807`. Another solution is https://github.com/Kong/kong-build-tools/pull/462 .


chronolaw (approved):
LGTM.

PR #468, opened 2022-05-18
FTI-3275 partially revert d66b848a b50c155b

outsinre:
We revert the `kong:kong` to `kong:root` change  at https://github.com/Kong/kong-build-tools/pull/457 and https://github.com/Kong/kong-build-tools/pull/463 to keep consistent with https://github.com/Kong/kong-distributions/pull/783.

The suggested files permission should be left alone as `root:root`. We should do in separate PRs to do this.


outsinre:
duplicate.

PR #471, opened 2022-05-23
fix(hybrid): fix the path to the wrpc file

hutchic:
tested as follows:
```
make package-kong
dpkg -c output/kong-2.8.0.20.04.amd64.deb | grep wrpc.proto
-rw-r----- 0/0            6271 2022-05-23 13:21 ./usr/local/kong/lib/include/wrpc/wrpc.proto
```

a backwards compatible tests would be a quick smoke of running hybrid Kong that I don't have the bandwidth to take on but it's on the todo list

### DIFF
--- a/build-kong.sh
+++ b/build-kong.sh
@@ -24,8 +24,7 @@ export PATH=$PATH:/usr/local/openresty/luajit/bin
 /usr/local/openresty/bin/openresty -v
 
 pushd /kong
-  cp -r kong/include/kong /tmp/build/usr/local/kong/lib/ || true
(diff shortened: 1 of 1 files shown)


bungle (approved):

zhongchen:
https://github.com/Kong/kong-ee/blob/d6d7b32cab7dda0819e18f1635febf1c43f80638/kong/tools/wrpc.lua#L124 is the file paths where wrpc search for `proto` definitions. I actually don't understand how your changes fix the issue.

PR #475, opened 2022-05-25
fix(tests): account for eventual consistency

jschmid1:
Signed-off-by: Joshua Schmid <jaiks@posteo.de>


bungle (approved):

jschmid1:
updated with a do-loop that isn't blocking on first success

PR #477, opened 2022-05-27
chore(release): adjust kong docker image releases

hutchic:
related: https://github.com/Kong/kong-ee/pull/3093

**Previously**
- we only pushed alpine as `namespace/image:x.y.z` ie `kong/kong:1.2.3`
- we pushed a `latest` moving tag

**Now**
- don't tag a `latest`
- we now push `namespace/image:x.y.z-distroversion` ie `kong/kong:1.2.3-alpine3` and `kong/kong:1.2.3-debian7`


hutchic:
I've reconsidered

PR #484, opened 2022-06-07
feat(*) We are bundling our proto files into the Kong luarock 

tyler-ball:
We no longer need to copy the 'include' folder to our build directory.
Instead we let Luarocks handle including these proto files (and any
future include files).

I also switched the protobuf version to be read from the `.requirements` file since Google regularly releases it.


curiositycasualty:
obsolete

PR #485, opened 2022-06-09
feat(cache-from): use docker buildkit cache-from ability

hutchic:
utilize docker buildkit `--cache-from` so we get higher fidelity than our current cache hit which is either the entire image already exists or it does not

Before
image

After
image


curiositycasualty (approved):
This looks good. I'd love to see some numbers about what this saves us on build times.

PR #492, opened 2022-06-14
chore(ci): add kong-ci-cache-key

hutchic:
We desire the ability to have a more local faster cache in places. To avoid side effects of overlapping cache's that may or may not hit / refresh on different requirements let's share a central unified authoritative cache key from one place.

An example of the cache key in use https://github.com/Kong/kong-ee/pull/3322/commits/866f3b18ff9b949b0b733e755c920ca8ccfc6b95

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+$(info starting make in kong-build-tools)
+
 VARS_OLD := $(.VARIABLES)
 
 .PHONY: test build-kong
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,5 +1,3 @@
-$(info starting make in kong-build-tools)
-
```
hutchic, line 2: @curiositycasualty I want this back. I couldn't find a way to suppress it's output conditionally -- any help appreciated :)

hutchic:
found a workaround

PR #501, opened 2022-06-29
fix(per-commit) add a `nightly` label

gszr:
Historically, kong-build-tools would update the `latest` tag
for each new per-commit release. Since acfda16, that is no longer the
case. 

Given it's useful to have a generic name for the latest build of
kong - and we want to avoid pushing a tag named `latest`, so as to avoid
surprises when users pull `kong/kong`- this commit adds a `nightly` tag.
Such tag will be used, for example, in decK's CI, where we want to test
against the latest available build of Kong.

Following this PR, the `latest` tag should also be removed from `kong/kong`
(it is still pointing to an old per-commit release).

(I am open to suggestions for a better name, since `nightly` doesn't really
represent the periodicity of the build -- it's generated per commit pushed to
Kong's `master` branch.)


gszr:
Perhaps `kong/kong:master` instead of `kong/kong:nightly`?

GGabriele:
+1 to `kong/kong:master` for me. That's more descriptive and gives a better idea of "what" and "where", instead of a misleading "when" for `nightly`.

aboudreault:
looks similar to https://github.com/Kong/kong/pull/9014?

hutchic:
adding onto @aboudreault looks similar to https://github.com/Kong/kong-ee/pull/3432 :joy:

GGabriele:
So much parallel effort lol

So, what is the current status? As per the commit referenced by @aboudreault , does now `latest` matches the current `master` branch?

gszr:
@hutchic Any updates on this? Should we merge or close this out?

hutchic:
I think we no longer need this with https://github.com/Kong/kong-ee/pull/3650 and https://github.com/Kong/kong/pull/9168

gszr:
Closing this. Thanks all!

PR #504, opened 2022-07-11
fix(docker): consistently label all our docker images

hutchic:
kong-ee migrated from `make release` to `make package-kong && make build-test-container`. This led to breaking gojira because the resulting image lacked some labels.

This PR applies the label to `build-test-container` and makes it consistent across all docker images


hutchic:
I didn't mean to create this from my fork

PR #506, opened 2022-07-11
feat(enterprise): install the kong-licensing nginx module only if needed

aboudreault:
We introduced `ENABLE_KONG_LICENSING` in kong-ee to be able to not build the kong-licensing C library (local dev and test purpose). However, KBT is still installing the nginx module, which can unset the licensing env variables on errors: https://github.com/Kong/kong-licensing/blob/b6e4ca3ee3747d76c87a1c66825bf5dcb4fabdf3/ngx_module/ngx_license_validation_module.c

this is why some of our license tests were failing in EE.


aboudreault:
@hutchic @curiositycasualty friendly ping in case you have some free cycles today.

Makefile
```diff
@@ -189,6 +190,7 @@ else
 	--build-arg DOCKER_BASE_SUFFIX=$(DOCKER_BASE_SUFFIX) \
 	--build-arg LIBYAML_VERSION=$(LIBYAML_VERSION) \
 	--build-arg EDITION=$(EDITION) \
+	--build-arg ENABLE_KONG_LICENSING=$(ENABLE_KONG_LICENSING) \
```
hutchic, line 193: I'd have thought this is the only spot this is required in the `Makefile` no?

aboudreault: umm... good point. I guess EDITION is a bit different because it is not only used when building openresty. let me double check this.

aboudreault: fixed

hutchic (approved):

PR #510, opened 2022-07-15
chore(ci): adjustments to make the dev image gojira compatible refs: #ENGEN-709

hutchic:


aboudreault:
so with this change, gojira will call KBT as normal, instead of calling kong-ngx-build itself?

hutchic:
@aboudreault I'm not sure it's not a fully formed thought yet to be honest.

I drafted a RFC PR [here](https://github.com/Kong/gojira/pull/62/files) but in general the goal would be to unify anything within the scope of "we need to build Kong pre-requisites / Kong itself" to one tool

PR #517, opened 2022-07-25
chore(build) atc_router is main by default

chronolaw:
Now atc_router is `0` by default, it will not download atc_router and cause kong runtime error.

Though we can use env `ATC_ROUTER` in the command line but I think it should have a valid default value.


hutchic (approved):
In the future we should pin this to make sure backwards compatibility is maintained ie in a year from now when we need to hotfix 2.8 will this work?

chronolaw:
> In the future we should pin this to make sure backwards compatibility is maintained ie in a year from now when we need to hotfix 2.8 will this work?

I understand. If we consider older versions like 2.8, default `0` is reasonable, maybe we should pin it in the future.

PR #520, opened 2022-07-27
fix(arm64): use qemu to do the packaging step for arm64 images refs: #ENGEN-715

hutchic:
Goal of the PR is the addition of `release-kong-docker-images` make task that uses `build-test-container` to build production worthy docker images and push them (leaving the task of renaming `build-test-container` and a few misnamed environment variables for a subsequent PR)

Putting this capability into kong-build-tools as it's the sensible location to share the logic between Kong and Kong Enterprise without having to copy-pasta between the two repositories. The `build-test-container` image name change is backwards compatible as it was assigned a variable everywhere it was used.

Previous result
```
make package-kong && make build-test-container
localhost:3000/kong-ubuntu-20.04
```

Adjusted result
```
export DOCKER_RELEASE_REPOSITORY=kong/kong #default
make package-kong && make build-test-container
# locally available docker images
kong/kong:3.0.0-ubuntu-20.04 # multi-arch if building arm
kong/kong:amd64-3.0.0-ubuntu-20.04
kong/kong:arm64-3.0.0-ubuntu-20.04 # only if building arm
```

New behaviour
```
export ADDITIONAL_TAGS="hello world"
export DOCKER_RELEASE_REPOSITORY=foo/bar
make package-kong && make release-kong-docker-images
# pushed docker images
foo/bar:3.0.0-ubuntu-20.04 # multi-arch if building arm
foo/bar:amd64-3.0.0-ubuntu-20.04
foo/bar:arm64-3.0.0-ubuntu-20.04 # only if building arm

foo/bar:hello
foo/bar:amd64-hello
foo/bar:arm64-hello

foo/bar:world
foo/bar:amd64-world
foo/bar:arm64-world
```


Makefile
```diff
@@ -43,11 +43,12 @@ OFFICIAL_RELEASE ?= true
 PACKAGE_CONFLICTS ?= `grep PACKAGE_CONFLICTS $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 PACKAGE_PROVIDES ?= `grep PACKAGE_PROVIDES $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 PACKAGE_REPLACES ?= `grep PACKAGE_REPLACES $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
-DOCKER_RELEASE_REPOSITORY ?= `grep DOCKER_RELEASE_REPOSITORY $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
```
curiositycasualty, line 46: Did this get tested with kong-ee? I am skeptical this would correctly overwrite with the variable's value in `.requirements` (https://github.com/Kong/kong-ee/blob/c2d854b8dea021ab581588b8e758665a519e4746/.requirements#L9)

Makefile
```diff
@@ -410,34 +430,39 @@ ifneq ($(RESTY_IMAGE_BASE),src)
...
+	docker tag $(DOCKER_RELEASE_REPOSITORY):amd64-$(KONG_TEST_CONTAINER_TAG) \
+		$(DOCKER_RELEASE_REPOSITORY):$(KONG_TEST_CONTAINER_TAG)
+	docker tag $(DOCKER_RELEASE_REPOSITORY):amd64-$(KONG_TEST_CONTAINER_TAG) \
+		$(KONG_TEST_IMAGE_NAME)
 ifeq ($(BUILDX),true)
-	DOCKER_MACHINE_NAME=$(shell docker-machine env $(DOCKER_MACHINE_ARM64_NAME) | grep 'DOCKER_MACHINE_NAME=".*"' | cut -d\" -f2) \
-	DOCKER_TLS_VERIFY=1 \
-	DOCKER_HOST=$(shell docker-machine env $(DOCKER_MACHINE_ARM64_NAME) | grep 'DOCKER_HOST=".*"' | cut -d\" -f2) \
-	DOCKER_CERT_PATH=$(shell docker-machine env $(DOCKER_MACHINE_ARM64_NAME) | grep 'DOCKER_CERT_PATH=".*"' | cut -d\" -f2) \
+	docker run --rm --privileged multiarch/qemu-user-static --reset -p yes && \
```
curiositycasualty, line 452: So, does this PR do away with docker-machine? If so, we should also strip out the docker-machine install.

README.md
```diff
@@ -22,28 +22,19 @@ Packaging arm64 architectures additionally requires:
```
curiositycasualty, line 22: Thanks for updating the README!

test/tests/01-package/run.sh
```diff
@@ -8,7 +8,7 @@ then
   major="${RESTY_IMAGE_TAG%%.*}"
   docker run -d --name user-validation-tests --rm -e KONG_DATABASE=off -v $PWD:/src registry.access.redhat.com/ubi${major}/ubi tail -f /dev/null
 else
-  docker run -d --name user-validation-tests --rm -e KONG_DATABASE=off -v $PWD:/src ${RESTY_IMAGE_BASE}:${RESTY_IMAGE_TAG} tail -f /dev/null
+  docker run -d --name user-validation-tests --rm -e KONG_DATABASE=off -v $PWD:/src --platform linux/amd64 ${RESTY_IMAGE_BASE}:${RESTY_IMAGE_TAG} tail -f /dev/null
```
curiositycasualty, line 11: Does this mean we aren't tailoring the tests to their specific architectures?

hutchic: This is so when the host system is amd64 the build doesn't go
```
build amd64
build arm64
accidentally use arm64 for testing because it's what's laying around
```

Now that I type this out it's occurred to me this will fix the bug for me (amd64) but break for m2 (arm64) :(

hutchic:
```suggestion
  docker rmi ${RESTY_IMAGE_BASE}:${RESTY_IMAGE_TAG} --force
  docker run -d --name user-validation-tests --rm -e KONG_DATABASE=off -v $PWD:/src ${RESTY_IMAGE_BASE}:${RESTY_IMAGE_TAG} tail -f /dev/null
```

I think will fix the situation right?

Makefile
hutchic: It does not.

We use docker-machine / a remote arm worker for the purposes of building Kong arm64. The final step of packaing and placing said package onto the host system `output/` directory is via qemu.

Makefile
hutchic: I couldn't find a way to keep this PR backwards compatible with current Kong CE which doesn't have a value for `DOCKER_RELEASE_REPOSITORY`.

Someway of in order of precedence:
1. use an environment variable
2. use the DOCKER_RELEASE_REPOSITORY value from .requirements
3. some sensible default

curiositycasualty: I think our build infrastructure for the gateway product already presumes that default values pertain to the building of CE. So option 3, a sensible default for the value required for the CE build, makes sense to me.

PR #523, opened 2022-08-08
feat(make): Use a (Make) Macro for Requirements

curiositycasualty:
Centralizes all the reading from a `.requirements` file into a single Make macro `_req`.

The shape of the macro call:
```
`$(call _req,…)`
```

might look weird here, but it keeps the functionality the same, and more importantly, keeps the `debug:` target working.


curiositycasualty:
Closing own stale PRs.

PR #524, opened 2022-08-11
feat(fips) enable fips mode by default on boringssl build

gszr:


gszr:
The package naming change can be seen in effect here: https://internal.builds.konghq.com/blue/organizations/jenkins/kong-build-tools/detail/PR-524/10/pipeline#step-87-log-2235.

hutchic (approved):
seems fine. We should make the necessary changes such that we're building a per commit fips container (previously known as a "nightly")

curiositycasualty (approved):

gszr:
> seems fine. We should make the necessary changes such that we're building a per commit fips container (previously known as a "nightly")

Agreed. I will card that.

PR #528, opened 2022-08-26
chore(*) add desc for `--atc-router`

mayocream:


openresty-build-tools/kong-ngx-build
```diff
@@ -1042,6 +1042,9 @@ show_usage() {
   echo "      --resty-websocket <tag>        Specify a Kong/lua-resty-websocket tag to install in place of the"
   echo "                                     bundled version of lua-resty-websocket (Defaults to \"$RESTY_WEBSOCKET\")"
   echo ""
+  echo "      --atc-router <tag>             Specify a Kong/atc-router tag to install in place of the"
```
flrgh, line 1045: It looks like the default behavior is to not include Kong/atc-router at all ([link](https://github.com/Kong/kong-build-tools/blob/a28053c73243ad6d7b5e6d959b459186d015d60d/openresty-build-tools/kong-ngx-build#L442)), so I think the language here could be updated to reflect that (instead of `in place of the bundled version`).

mayocream: Thanks! I updated the desc now.

flrgh (approved):
:sunglasses:

PR #530, opened 2022-08-26
fix(release): 3.0 release fixes and improvements

tyler-ball:
Tested CE and EE release pipelines. The biggest change in this is the
introduction of the KONG_RELEASE_LABEL variable that replaces the
overloaded KONG_VERSION variable. This allows us to label a release with
a version read from meta.lua or a git tag.

Also includes fixes for the FIPS build pipeline, and assorted fixes we
came across in testing.


release-kong.sh
```diff
@@ -10,14 +10,14 @@ PULP_USERNAME=
 PULP_PASSWORD=
 
 # release finals into prod, others into stage
-if [[ "$KONG_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
```
tyler-ball, line 13: `KONG_VERSION` looks really overloaded. Right now my tag is `3.0.0.0-tyler` but `KONG_VERSION` is read from the ee files so comes in as `3.0.0.0` and fell into the 'prod' path of this tree.

Possibilities:
Depend more on `OFFICIAL_RELEASE` and a new `KONG_TAG` environment variable
Set a suffix in kong/enterprise_edition/meta.lua (I think this is how CE handles this)

Makefile
```diff
@@ -46,12 +46,18 @@ PACKAGE_REPLACES ?= `grep PACKAGE_REPLACES $(KONG_SOURCE_LOCATION)/.requirements
 DOCKER_RELEASE_REPOSITORY?="kong/kong"
 
 KONG_TEST_CONTAINER_NAME=kong-tests
-KONG_TEST_CONTAINER_TAG?=$(KONG_VERSION)-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)
+KONG_TEST_CONTAINER_TAG?=$(KONG_RELEASE_LABEL)-$(RESTY_IMAGE_BASE)-$(RESTY_IMAGE_TAG)
```
hutchic, line 49:
```suggestion
KONG_TEST_CONTAINER_TAG?=$(KONG_RELEASE_LABEL)-$(RESTY_IMAGE_BASE)
```
we don't desire the OS version appended to the docker tag

hutchic: replaces https://github.com/Kong/kong-build-tools/pull/537

release-kong.sh
tyler-ball: I can definitely see us coming back to address this later. Right now commits to `master` and `next/3.0.x.x` go to pulp prod, but if I tag `3.0.0.0-preview` from `next/3.0.x.x` it will go to pulp dev

hutchic (approved):
Make sure the comment message meets semantic commit including the colon

fpm-entrypoint.sh
```diff
@@ -40,8 +40,9 @@ then
   PACKAGE_REPLACES=kong-community-edition
   PACKAGE_REPLACES_2=kong-enterprise-edition-fips
 
-elif [ "$KONG_PACKAGE_NAME" = "kong-enterprise-edition-fips" ]
+elif [ "$KONG_PACKAGE_NAME" = "kong-enterprise-edition-fips" ] || [ "$SSL_PROVIDER" = "boringssl" ]
```
gszr, line 43:
```suggestion
elif [ "$SSL_PROVIDER" = "boringssl" ]
```
(the first clause will never be true, unless the package name is already set, in which case it's unnecessary to set it again)

hutchic: `KONG_PACKAGE_NAME` [should](https://github.com/Kong/kong-build-tools/blob/master/Makefile#L40) always be set

PR #537, opened 2022-08-29
fix(docker): don't include the OS version in the docker tag

hutchic:
before
```
kong/kong:ubuntu20.04
```

after
```
kong/kong:ubuntu
```


hutchic:
superseded by https://github.com/Kong/kong-build-tools/pull/530

PR #544, opened 2022-09-10
fix(release) Use the correct redhat docker registry format

tyler-ball:
Looking at https://catalog.redhat.com/software/containers/search it looks like all the different rhel distributions are at a top level name in the Red Hat registry, IE

registry.access.redhat.com/ubi7
registry.access.redhat.com/ubi8
registry.access.redhat.com/ubi8-minimal

I'm not sure where the 2nd part of the package name came from in the original sed.

Signed-off-by: Tyler Ball <tyler.ball@konghq.com>

### DIFF
--- a/test/build_container.sh
+++ b/test/build_container.sh
@@ -36,7 +36,7 @@ pushd ./docker-kong
   then
     major="${RESTY_IMAGE_TAG%%.*}"
 
-    sed -i.bak "s|^FROM .*|FROM registry.access.redhat.com/ubi${major}/ubi:${RESTY_IMAGE_TAG}|" Dockerfile.$PACKAGE_TYPE
+    sed -i.bak "s|^FROM .*|FROM registry.access.redhat.com/ubi${major}|" Dockerfile.$PACKAGE_TYPE
   elif [ "$RESTY_IMAGE_BASE" == "debian" ]; then
     sed -i.bak 's/^FROM .*/FROM '${RESTY_IMAGE_BASE}:${RESTY_IMAGE_TAG}-slim'/' Dockerfile.$PACKAGE_TYPE
   else


test/build_container.sh
```diff
@@ -36,7 +36,7 @@ pushd ./docker-kong
   then
     major="${RESTY_IMAGE_TAG%%.*}"
 
-    sed -i.bak "s|^FROM .*|FROM registry.access.redhat.com/ubi${major}/ubi:${RESTY_IMAGE_TAG}|" Dockerfile.$PACKAGE_TYPE
+    sed -i.bak "s|^FROM .*|FROM registry.access.redhat.com/ubi${major}|" Dockerfile.$PACKAGE_TYPE
```
gszr, line 39: @curiositycasualty Isa, do you know why this started failing recently? Seems it did work in the PR that changed it.

hutchic: The problem is detailed on the PR that added this change. RHEL doesn't provide floating container at 
`registry.access.redhat.com/ubi7` but instead requires the minor to be pinned `registry.access.redhat.com/ubi7.9`

This explains why CE didn't require this change but EE did.

CE is explicitely pinning the rhel [major.minor 7.9 and 8.6](https://github.com/Kong/kong/blob/master/Jenkinsfile#L137-L138)
EE attempted to use a floating tag of [major 7 and 8](https://github.com/Kong/kong-ee/blob/3.0.0.0/Jenkinsfilev2#L175-L176)

If we could get the best of both that would be ideal -- pin major.minor but if it's forgotten still be backwards compatible :thinking:

gszr: We will pin to major.minor in kong-ee master and revisit this change later.

hutchic (approved):

test/build_container.sh
tyler-ball: Why wouldn't we always want the latest version for a given major, except except to avoid a new CVE or something similar? I'm proposing we default to float on `latest`

PR #548, opened 2022-09-13
refactor(fpm): cleanup fpm tooling

flrgh:
This is a sizeable refactor of the fpm entrypoint script. The goal of this is to improve reliability, debug-ability, and readability.

* fix some shell scripting gotchas (unquoted args, relying on potentially undefined vars, etc)
* normalize some variable names so they make more sense in the package-building context
* flatten all of the business logic into `case` blocks rather than lots of changed `if`/`elif`/`else`
* Dockerfile.package updates
  - remove duplicate build arg
  - add DEBUG build-arg/env var
  - fixed line endings (`dos2unix`); this make the diff completely useless though :(

It's a pretty simple script--conditionally set some vars and then run one of two commands to generate a package--so I stopped short of breaking things into functions and whatnot. Nevertheless, it's a large change, so step through it commit-by-commit if you want a better picture of what all has changed and why.

~There are still a couple things I want to tidy up, and I haven't tested all of the build permutations locally yet, so this is dubbed WIP, but feel free to offer feedback right now.~ I've done basic testing (`make package-kong` and installing the generated package in a container) with each supported distro. Only one that isn't working is alpine, but I think that one requires some extra steps anyhow (since it fails in the same way on master for me).


dockerfiles/Dockerfile.package
```diff
@@ -1,50 +1,51 @@
...
-
-RUN mkdir -p /tmp/build/lib/systemd/system/
-COPY kong.service /tmp/build/lib/systemd/system/kong.service
-COPY kong.logrotate /tmp/build/etc/kong/kong.logrotate
-
-RUN /fpm-entrypoint.sh
-
-FROM alpine
-COPY --from=FPM /output /output
+ARG PACKAGE_TYPE
```
flrgh, line 1: To reviewers: this file was corrected from DOS to Unix-style line-endings, so the diff here is useless. Changes are annotated in my comments.

dockerfiles/Dockerfile.package
```diff
@@ -1,50 +1,51 @@
...
+ENV PRIVATE_KEY_PASSPHRASE ${PRIVATE_KEY_PASSPHRASE}
+
+ARG RESTY_IMAGE_BASE="ubuntu"
+ARG RESTY_IMAGE_TAG="18.04"
+
+ARG PACKAGE_TYPE
+ARG KONG_VERSION
+ARG KONG_RELEASE_LABEL
+ARG KONG_PACKAGE_NAME
+ARG TARGETPLATFORM
```
flrgh, line 26: Removed the duplicate `TARGETPLATFORM` build arg/env var.

dockerfiles/Dockerfile.package
```diff
@@ -1,50 +1,51 @@
...
+ARG PACKAGE_PROVIDES="kong-community-edition"
+ENV PACKAGE_PROVIDES $PACKAGE_PROVIDES
+
+ARG PACKAGE_REPLACES="kong-community-edition"
+ENV PACKAGE_REPLACES $PACKAGE_REPLACES
+
+ARG SSL_PROVIDER
+ENV SSL_PROVIDER $SSL_PROVIDER
+
+ARG DEBUG=0
```
flrgh, line 41: Added the `DEBUG` build arg/env var.

PR #557, opened 2022-10-11
chore(release) Deduplicating grep-kong-version.sh script

tyler-ball:
Currently this script lives in both the kong and kong-ee repos, but we only use it in those repos to pass the Kong version to KBT.

Fixes FT-3166

Signed-off-by: Tyler Ball <tyler.ball@konghq.com>


Makefile
```diff
@@ -45,8 +45,7 @@ PACKAGE_PROVIDES ?= `grep PACKAGE_PROVIDES $(KONG_SOURCE_LOCATION)/.requirements
 PACKAGE_REPLACES ?= `grep PACKAGE_REPLACES $(KONG_SOURCE_LOCATION)/.requirements | awk -F"=" '{print $$2}'`
 DOCKER_RELEASE_REPOSITORY?="kong/kong"
 
-# This logic should mirror the kong-build-tools equivalent
-KONG_VERSION?=`$(KONG_SOURCE_LOCATION)/distribution/grep-kong-version.sh`
+KONG_VERSION?=`./grep-kong-version.sh $(KONG_SOURCE_LOCATION)`
```
flrgh, line 48: Since this should only live in this repo from now on, WDYT about renaming this to `detect-kong-version` or something like that?

grep-kong-version.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# unofficial strict mode
+set -euo pipefail
+
+KONG_SOURCE_LOCATION=$1
+
+kong_version=`echo $KONG_SOURCE_LOCATION/kong-*.rockspec | sed 's,.*/,,' | cut -d- -f2`
+
+if test -f "$KONG_SOURCE_LOCATION/kong/enterprise_edition/meta.lua"; then
+    ee_patch=`grep -o -E 'ee_patch[ \t]+=[ \t]+[0-9]+' $KONG_SOURCE_LOCATION/kong/enterprise_edition/meta.lua | awk '{print $3}'`
```
flrgh, line 11:
```suggestion
    ee_patch=$(grep -o -E 'ee_patch[ \t]+=[ \t]+[0-9]+' "$KONG_SOURCE_LOCATION"/kong/enterprise_edition/meta.lua | awk '{print $3}')
```

grep-kong-version.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
flrgh, line 1: This script is in pretty rough shape, but I get that this is not what we're trying to fix at the moment. I'm suggesting a couple small improvements:

1. use `$(command)` instead of backticks
2. quote usage of `$KONG_SOURCE_LOCATION`

flrgh (approved):
I left some suggestions inline, but it's nothing that can't be done later on.

dndx (approved):

PR #558, opened 2022-10-13
fix(*) drop RHEL 7

fffonion:
Reflect the product update for RHEL 7 sunset.

For FTI-4303, DOCU-2489


fffonion:
Question: do we need to update Jenskinfile in kong repos?

PR #566, opened 2022-10-21
fix(arm): pin the base image so arm builds succeed

hutchic:
just getting the problem fixed. I'll come back to `--platform` flags and an earlier / more obvious test later

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -139,6 +139,7 @@ pipeline {
                     environment {
                         PATH = "/home/ubuntu/bin/:${env.PATH}"
                         PACKAGE_TYPE = "rpm"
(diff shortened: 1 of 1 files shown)


dockerfiles/Dockerfile.openresty
```diff
@@ -2,7 +2,7 @@ ARG DOCKER_BASE_SUFFIX
 ARG DOCKER_REPOSITORY
 ARG PACKAGE_TYPE
 
-FROM kong/kong-build-tools:$PACKAGE_TYPE
```
hutchic, line 5: dependabot would be handy here but I'm not sure how to achieve this

gszr (approved):

PR #574, opened 2022-10-25
fix(after-install) do not create home

gszr:
This also adds a comment (GEOOS field) to the Kong user.


hutchic (commented):
if @outsinre says it's fine I think it's fine

fffonion (approved):

PR #575, opened 2022-10-27
tests(systemd): wait for nginx pidfile

flrgh:
We have had some CI wonkiness where the `systemctl reload kong` package test fails due to an empty pid file:

> Oct 26 21:20:27 0f37d35f7c2e nginx[607]: nginx: [error] invalid PID number "" in "/usr/local/kong/pids/nginx.pid"

Seems like nginx was just taking slightly longer to start than expected, so I've made the following changes:

1. added a wait loop that will block until the nginx pidfile is non-empty (15s max). It should be a little more reliable than the prior approach of sleeping for 5s and continuing on.
2. pinned the worker process count to 2, as it should make for a consistently faster startup time.


hutchic (approved):
oooh that's much better than randomly sleeping 5 seconds. Thank you

PR #585, opened 2022-11-03
fix(packaging): we don't want Kong to replace kong-enterprise-edition refs: #INC-155

hutchic:
Fixes an issue where the 3.x yum/rpm repos were installing (CE) kong when asked to install `kong-enterprise-edition`.


curiositycasualty (approved):

tyler-ball (approved):
Note for future us: we removed `PACKAGE_REPLACES_2` so we went back to the 2.8 behavior. In 3.1 we should decide how we want to handle the CE/EE/FIPS editions and their `REPLACES` hierarchy

gszr (approved):

PR #589, opened 2022-11-04
fix(systemd): migrate the systemd test from ubuntu to rhel

hutchic:


test/tests/01-package/run.sh
```diff
@@ -181,44 +181,38 @@ if [ "$SSL_PROVIDER" = "openssl" ]; then
...
-  docker exec ${USE_TTY} systemd-ubuntu /bin/bash -c "apt install --yes /src/kong.deb"
-  docker exec ${USE_TTY} systemd-ubuntu /bin/bash -c "kong version"
-
-  docker exec ${USE_TTY} systemd-ubuntu /bin/bash -c "test -f /etc/kong/kong*.logrotate"
-  docker exec ${USE_TTY} systemd-ubuntu /bin/bash -c "mkdir -p /etc/systemd/system/kong.service.d/"
-  docker exec ${USE_TTY} systemd-ubuntu /bin/bash -c "cat <<\EOD > /etc/systemd/system/kong.service.d/override.conf
+if [[ "$RESTY_IMAGE_BASE" == "rhel" ]] && [ -z "${DARWIN:-}" ]; then
+  cp $PACKAGE_LOCATION/*${KONG_ARCHITECTURE}.rpm kong.rpm
+  #docker run ${USE_TTY} --name systemd-ubuntu -e KONG_DATABASE=off --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v $PWD:/src jrei/systemd-ubuntu:$RESTY_IMAGE_TAG
+  docker run -d --rm --name=systemd --privileged --tmpfs /tmp --tmpfs /run --tmpfs /run/lock -v $PWD:/src redhat/ubi8-init
```
hutchic, line 187: `redhat/ubi8-init` with something ....

curiositycasualty (approved):

test/tests/01-package/run.sh
hutchic: fixed by only running the tests when the build is rhel 8. So systemd is tested for all rpm's but not for our deb but since it's the same systemd unit file that **should** be fine ....

PR #607, opened 2022-11-17
chore(atc-router): capability of using a pre-compiled atc-router

hutchic:
Waiting on https://github.com/Kong/atc-router/pull/15 but once that's merged / released going forward we should use pre-built ATC router


hutchic:
replaced by https://github.com/Kong/kong-build-tools/pull/608

PR #610, opened 2022-11-22
chore(docker-images): add fips suffix if boringssl

gszr:


gszr:
We don't need this for now.

PR #627, opened 2022-12-11
fix(arm64): shrink cargo mem footprint

curiositycasualty:
The arm64 build was relatively silently failing while either doing the `cargo build` or just fetching the cargo registry index.
This resolves that issue by taking steps to minimize `cargo`'s memory footprint when building on arm64.

It also splits the `atc-router` install across the openresty build. Building `atc_router.so` before openresty, but __installing__ `atc_router.so` __after__ openresty-- since the openresty files themselves are not strictly required to build atc-router, just for placing the .so in the correct directory (as done by `make`).

Building atc-router sooner in build helps us fail faster when something with the atc-router goes awry.

### DIFF
--- a/openresty-build-tools/kong-ngx-build
+++ b/openresty-build-tools/kong-ngx-build
@@ -482,6 +482,42 @@ main() {
     exit 0
   fi
 
+  if [ $ATC_ROUTER != 0 ]; then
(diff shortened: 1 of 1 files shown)


curiositycasualty:
Closing own stale PRs.

PR #629, opened 2022-12-13
fix(arm64,luajit): bad ptr address with debug mode

pluveto:
We (@dndx ) found that for every Mac M1 aarch64 device (running MacOS or Ubuntu) after building LuaJIT using command such as 
```
./kong-ngx-build -p buildroot --openresty 1.21.4.1 --openssl 1.1.1s --pcre 8.45 -f --debug
```

the LuaJIT binary cannot init properly, which prints

```
./buildroot/openresty/luajit/bin/luajit: cannot create state: not enough memory
```

(this can be reproduced every time)

and @dndx found if `LUAJIT_USE_SYSMALLOC` is enabled in that platform (which is enabled while adding `--debug` flag), system malloc function (`realloc` virtually) returns a address pointer, of which 47 bit is high, while LuaJIT uses this high bit to express some metadata so that this address failed to pass LuaJIT's address check. So, LuaJIT's solution (https://github.com/LuaJIT/LuaJIT/issues/49) is to use `mmap` which allow us to specific an address that high bits are zero.

So we removed `DLUAJIT_USE_SYSMALLOC` when the architecture is `aarch64` to ensure LuaJIT get a legal address.

### DIFF
--- a/openresty-build-tools/kong-ngx-build
+++ b/openresty-build-tools/kong-ngx-build
@@ -841,7 +841,11 @@ main() {
         if [ $DEBUG == 1 ]; then
           OPENRESTY_OPTS+=('--with-debug')
           OPENRESTY_OPTS+=('--with-no-pool-patch')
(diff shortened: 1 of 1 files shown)


fffonion:
@pluveto Could you reopen this PR as a branch instead of a fork? I think the CI doens't run on PR and I can't merge it.

PR #644, opened 2023-09-05
Fix dependencies submodule initialization

Elyahou:
Until now, the submodules of dependencies are fetched at the clone time, and are nor updated after changing branch to the required version. This lead to wrong versions of submodules being used and cause issues.

Fix:
Initialize submodules after changing branch to use submodules version of the required branch.


hanshuebner:
This repository is no longer maintained as Kong is now built by Bazel: https://github.com/Kong/kong/tree/master/build

